PR #1, opened 2019-11-18
Add build pipeline

sjmiller609:
Thanks Greg for getting this repo ready!

epic: https://github.com/astronomer/issues/issues/465

A push to any branch will publish:
astronomerinc/ap-airflow:1.10.5-alpine-onbuild
astronomerinc/ap-airflow:1.10.5-alpine3.10-onbuild
astronomerinc/ap-airflow:1.10.5-buster-onbuild

A push to master will publish:
astronomerinc/ap-airflow:1.10.5-alpine
astronomerinc/ap-airflow:1.10.5-alpine3.10
astronomerinc/ap-airflow:1.10.5-buster

Is this as intended?

All images are only published if they pass the Clair scan for high-severity issues.

Pre-merge checklist:
- [x] Circle CI: Turn on for this repo
- [x] Circle CI: Opt-in to third party orbs
- [x] Circle CI: Add DOCKER_USERNAME DOCKER_PASSWORD to circle CI environment variables

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,14 +15,14 @@ jobs:
     executor: docker-executor
     steps:
       - push:
-          comma_separated_tags: "1.10.5-alpine-onbuild,1.10.5-alpine3.10-onbuild"
+          comma_separated_tags: "1.10.5-alpine3.10-$CIRCLE_SHA1,1.10.5-alpine3.10-$CIRCLE_BUILD_NUM"
           image_name_load: ap-airflow-alpine
           image_name_publish: ap-airflow
   publish-prod-alpine:
     executor: docker-executor
     steps:
       - push:
-          comma_separated_tags: "1.10.5-alpine,1.10.5-alpine3.10"
+          comma_separated_tags: "1.10.5-alpine,1.10.5-alpine3.10,1.10.5-alpine-onbuild,1.10.5-alpine3.10-onbuild"
           image_name_load: ap-airflow-alpine
           image_name_publish: ap-airflow
   build-buster:
@@ -40,14 +40,14 @@ jobs:
     executor: docker-executor
     steps:
       - push:
-          comma_separated_tags: "1.10.5-buster-onbuild"
+          comma_separated_tags: "1.10.5-buster-$CIRCLE_SHA1,1.10.5-buster-$CIRCLE_BUILD_NUMBER"
           image_name_load: ap-airflow-buster
           image_name_publish: ap-airflow
   publish-prod-buster:
     executor: docker-executor
     steps:
       - push:
-          comma_separated_tags: "1.10.5-buster"
+          comma_separated_tags: "1.10.5-buster,1.10.5-buster-onbuild"
           image_name_load: ap-airflow-alpine
           image_name_publish: ap-airflow
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,168 @@
+version: 2.1
+jobs:
+  build-alpine:
+    executor: docker-executor
+    steps:
+      - docker-build:
+          image_name: ap-airflow-alpine
+          path: 1.10.5/alpine3.10
+  scan-alpine:
+    executor: clair-scanner/default
+    steps:
+      - clair-scan:
+          image_name: ap-airflow-alpine
+  publish-dev-alpine:
+    executor: docker-executor
+    steps:
+      - push:
+          comma_separated_tags: "1.10.5-alpine-onbuild,1.10.5-alpine3.10-onbuild"
+          image_name_load: ap-airflow-alpine
+          image_name_publish: ap-airflow
+  publish-prod-alpine:
+    executor: docker-executor
+    steps:
+      - push:
+          comma_separated_tags: "1.10.5-alpine,1.10.5-alpine3.10"
+          image_name_load: ap-airflow-alpine
+          image_name_publish: ap-airflow
+  build-buster:
+    executor: docker-executor
+    steps:
+      - docker-build:
+          image_name: ap-airflow-buster
+          path: 1.10.5/buster
+  scan-buster:
+    executor: clair-scanner/default
+    steps:
+      - clair-scan:
+          image_name: ap-airflow-buster
+  publish-dev-buster:
+    executor: docker-executor
+    steps:
+      - push:
+          comma_separated_tags: "1.10.5-buster-onbuild"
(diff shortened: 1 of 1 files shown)


schnie:
This one is a little different and mimics how popular library images (python, node, etc) are structured. This is one case where I think we may want to re-push tags if we introduce a fix or patch. Since we're now versioning the airflow tags using the `airflowVersion` + `baseDistro`, there's nothing to increment if we introduce a patch. If we introduce a patch, I think we'll want to _update_ the existing tag with the patched version.

If we only update one specific image (imagine a larger version + distro matrix), we really don't have to push tags for every image in this repo - only the one that was updated. This seems to make it a more manual / intentional kind of thing.

Does circle have similar functionality as drone with `drone deploy` / `drone promote`, where you can pass args for the build? Thinking we could specify a version/distro and have it automatically build and push that tag or something.

Or we could have it similar to our other repos, where we cut a github release, which triggers a production pipeline to build and push all tags, which maybe isn't the _worst_ thing.

To expand on this and clarify your original question @sjmiller609, the `onbuild` version of any docker image will need to be rebuilt and published if the parent image is updated. So say we update `1.10.5-alpine3.10`, then we'll need to also build and push `1.10.5-alpine3.10-onbuild`.

The main systematic tags we'll need to produce are `<airflowVersion>-<distro>` and `<airflowVersion>-<distro>-onbuild`, and any other tags we produce will really just be aliases to the other tags... or example:

- `1.10.5-alpine` would just be an easier way to say "give me the latest alpine version you have for airflow `1.10.5`, and would be an alias to `1.10.5-alpine3.10` currently. 
- `1.10.5` could be an easier way to say "give me the latest preferred airflow `1.10.5` image. This could point to either alpine or buster, whichever we want to be our default.

Once we have this sorted out, we can update our `ap-airflow` documentation page to look more similar to pythons - https://hub.docker.com/_/python/.

@sjmiller609 @ashb @vikramkoka any thoughts?

schnie:
Looking at https://hub.docker.com/_/python/?tab=tags, it seems like these library images take the approach of updating them all at once, just based on the "last updated" timestamps here.

schnie:
So it'd be cool to be able to cut a release on the repo and have circle parallelize a matrix build of `airflowVersion`/`distro` images, then push them up.

ashb:
I'm okay with repushing the "non-sepcific" tags, but it would also be nice to have a tag that includes something like build or release version so that it's possible to update or pin to a specific release.

So `1.10.5-alpine3.10-1` == `1.10.5-alpine3.10` == `1.10.5-alpine`

And then the next build would be `1.10.5-alpine3.10-2` == `1.10.5-alpine3.10` == `1.10.5-alpine` (but `1.10.5-alpine3.10-1` still exists if anyone needs it)

schnie:
Ah, maybe another optional portion of the tag to include the tag that we specify when we cut the release?

Maybe our first "production" release of this repo should wipe clean at `1.0.0` (independent of the platform or anything), and produce (alpine for example), `1.0.0-1.10.5-alpine3.10`, `1.10.5-alpine3.10`, `1.10.5-alpine`, `1.10.5`.

And maybe we just don't document the ones prefixed by the release version or something.

sjmiller609:
I'm addressing a lot of feedback here, so sorry for the long post! Summary at bottom.

@schnie 
> If we introduce a patch, I think we'll want to _update_ the existing tag with the patched version.

Ok, that's how this is now. Also, that's how all the other pipelines are (since I didn't do the versioning part yet)

> If we only update one specific image...
> 
> ... Does circle have similar functionality as drone with `drone deploy` / `drone promote`, where you can pass args for the build? Thinking we could specify a version/distro and have it automatically build and push that tag or something.

I could add logic in one of the jobs to exit if there is no change in the applicable docker directory since the last time. If we do that in the publish step only, then we will be alerted when clair scan fails, even for the unchanged images. If we do that for all jobs, then it will effectively skip the pipeline for unchanged docker directories.

> say we update `1.10.5-alpine3.10`, then we'll need to also build and push `1.10.5-alpine3.10-onbuild`.

Is there any situation where we want these two tags to have different content? If not, then I can just move all tags to the "prod" publishing steps.

What you said about 1.10.5-alpine pointing to the most recent version of alpine makes sense to me.

> So it'd be cool to be able to cut a release on the repo and have circle parallelize a matrix build of `airflowVersion`/`distro` images, then push them up.

Currently, the build will be parallel, but we are not using a Jenkins-style matrix build feature at the moment - each image is requiring us to define the jobs for it (but we are re-using the logic between then, so it's not that bad of duplication).

The build, scan, and publish workflow as-is in this PR is performed on a push to the master branch, rather than a release. If a PR is merged to master of this repo, then soon after customers will be able to use the new tags

@ashb 
> I'm okay with repushing the "non-sepcific" tags, but it would also be nice to have a tag that includes something like build or release version so that it's possible to update or pin to a specific release.
> 
> So `1.10.5-alpine3.10-1` == `1.10.5-alpine3.10` == `1.10.5-alpine`

This makes sense to me as well!

greg said:
> Maybe our first "production" release of this repo should wipe clean at `1.0.0` (independent of the platform or anything), and produce (alpine for example), `1.0.0-1.10.5-alpine3.10`, `1.10.5-alpine3.10`, `1.10.5-alpine`, `1.10.5`.

I could just publish the git commit hash as a label. That serves to perfectly track an image back to the code version. I can also do any combination of these suggestions. One thing I suggest is to omit an additional semantic versioning in the label because it is confusing in combination with the airflow version, and it does not have a clearly defined major/minor version increment objective. I would rather just use the Circle CI build number, which will be an increasing, but non-consecutive positive integer, and structure it as ash suggested.

Summary:
- which steps do we want to run, if any, on a push to master of this repo for an unchanged docker directory? steps (aka jobs) are build, security scan, publish
- do we want the publish step to only run on a release? The upside is a documented moment of release, the downside is an additional decision point that is similar in concept to the PR review and merge process.
- Is there any situation where X and X-onbuild are different?
- should I add X-<build number> to prod publish step?
- I'll add in git hash publish step to all branches

sjmiller609:
With the change I just pushed, now
```
1.10.5-alpine3.10-<git commit hash>
```
and
```
1.10.5-alpine3.10-<circle CI build number>
```
are publish on a push to any branch. On a push to master, these are published:
```
1.10.5-alpine
1.10.5-alpine3.10
1.10.5-alpine-onbuild
1.10.5-alpine3.10-onbuild
```
same idea for buster images.

PR #4, opened 2019-11-21
Circle image tags

schnie:
@sjmiller609 commenting out the dev ones for now. It kind of clutters up the repo. I want to think about those some more.

Also, the `onbuild` images are actually separate images, not additional tags on the original images. I've made some changes, but am hitting an issue with these builds. Trying to run the `onbuild` step right after the base image is built in both distros. For some reason its failing.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -61,8 +61,8 @@ jobs:
       - setup_remote_docker:
           docker_layer_caching: true
       - docker-build:
-        image_name: ap-airflow-buster
-        path: 1.10.5/buster
+          image_name: ap-airflow-buster
+          path: 1.10.5/buster
       - run:
           name: "Tag image"
           command: docker tag ap-airflow-buster astronomerinc/ap-airflow:1.10.5-buster
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -80,9 +80,9 @@ workflows:
       - scan-alpine:
           requires:
             - build-alpine
-      - publish-dev-alpine:
-          requires:
-            - scan-alpine
+      # - publish-dev-alpine:
+      #     requires:
+      #       - scan-alpine
       - publish-prod-alpine:
           requires:
             - scan-alpine
@@ -93,9 +93,9 @@ workflows:
       - scan-buster:
           requires:
             - build-buster
-      - publish-dev-buster:
-          requires:
-            - scan-buster
+      # - publish-dev-buster:
+      #     requires:
+      #       - scan-buster
       - publish-prod-buster:
           requires:
             - scan-buster
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,9 +6,14 @@ jobs:
   build-alpine:
     executor: docker-executor
     steps:
+      - checkout
+      - setup_remote_docker
       - docker-build:
           image_name: ap-airflow-alpine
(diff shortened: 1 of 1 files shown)


sjmiller609:
The error circle was reporting was because we tried to do setup_remote_docker twice in one step list. I moved that part outside of the docker-build command. I also added tagging at the right time so that the onbuild will build from the expected base image.

.circleci/config.yml
```diff
@@ -1,54 +1,86 @@
 version: 2.1
 jobs:
+  #
+  # Alpine
+  #
   build-alpine:
     executor: docker-executor
     steps:
+      - checkout
+      - setup_remote_docker
```
sjmiller609: this block shouldn't be repeated in a step list.

.circleci/config.yml
```diff
@@ -1,54 +1,86 @@
...
       - docker-build:
           image_name: ap-airflow-alpine
           path: 1.10.5/alpine3.10
+      - run:
+          name: "Tag image"
+          command: docker tag ap-airflow-alpine astronomerinc/ap-airflow:1.10.5-alpine3.10
```
sjmiller609: this block required to have onbuild build from local image instead of pulling

sjmiller609:
@schnie ok it's working now.

workflow now looks like this:

for all branches, at the same time:
- build alpine, then build alpine-onbuild. when done, clair scan each in parallel
- build buster, then build buster-onbuild. when done clair scan each in parallel

for master branch also do these two steps at the same time:
- if alpine scan passes on both alpine and alpine-onbuild, then publish both
- if buster scan passes on both buster and buster-onbuild, then publish both

also, docker layer caching is enabled, so if you make the Dockerfiles in a way where a local build can be re-built quickly on a typical change, then it will not take as long in the pipeline either. Consider the build number or other args that might effect caching.

.circleci/config.yml
schnie: @sjmiller609 instead of this extra step, can we change the `docker-build`'s `image_name` to be the actual desired name? Like `ap-airflow:1.10.5-alpine3.10` instead of `ap-airflow-alpine`.

.circleci/config.yml
```diff
@@ -28,38 +32,45 @@ jobs:
...
+  # Build Debian Images
   build-buster:
     executor: docker-executor
     steps:
       - checkout
       - setup_remote_docker:
           docker_layer_caching: true
       - docker-build:
-          image_name: ap-airflow-buster
+        image_name: ap-airflow-buster
```
sjmiller609: missing indent

schnie: incoming

sjmiller609:
I'm almost done with adding the testing part to the pipeline as well. that'll be in a different PR and include these changes.

schnie:
Nice, 5min for build scan both. :+1:

.circleci/config.yml
sjmiller609: there is an issue where the pathing is weird when the name contains a slash (for the save and load from workspace part). we can probably work around that somehow. you are saying without the organization name in it, so that might just work as you are suggesting. I'm good with whatever assuming the pipeline works, but always be cautious for accidentally pulling old images

sjmiller609:
It will take longer of course when there are changes

PR #7, opened 2019-11-21
Add testing for Airflow images

sjmiller609:
Adding testing for Airflow images. These will run at the same time as the clair scan.

It uses the following tooling:
- pytest
- docker
- testinfra

it does so like this:
- start postgres and airflow webserver (from image) using docker
- use pytest to run tests against the live, running container
- use testinfra for utility functions and syntactic sugar
example: 
the utility functions for configuration testing are provided by testinfra
```
def test_airflow_in_path(webserver):
    assert webserver.exists('airflow'), \
        "Expected 'airflow' to be in PATH"
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,6 +18,16 @@ jobs:
       - docker-build:
           image_name: ap-airflow-alpine-onbuild
           path: 1.10.5/alpine3.10/onbuild
+  test-alpine:
+    executor: docker-executor
+    steps:
+      - airflow-image-test:
+          image_name: ap-airflow-alpine
+  test-alpine-onbuild:
+    executor: docker-executor
+    steps:
+      - airflow-image-test:
+          image_name: ap-airflow-alpine-onbuild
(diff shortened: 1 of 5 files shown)


sjmiller609:
we might want to remove scan and test steps for non-onbuild images since the onbuild is layered on top of the other image, so it seems a bit redundant.

schnie:
Nice, looking good. Would be nice to see some comments that document what's happening in English and that can set the standard for other folks to add functionality.
https://github.com/astronomer/astronomer-fab-securitymanager/blob/master/src/astronomer/flask_appbuilder/security.py is a good example.

sjmiller609:
> Nice, looking good. Would be nice to see some comments that document what's happening in English and that can set the standard for other folks to add functionality.
> https://github.com/astronomer/astronomer-fab-securitymanager/blob/master/src/astronomer/flask_appbuilder/security.py is a good example.

sure, you mean in the test code, right?

PR #12, opened 2019-12-04
Fix buster image

andriisoldatenko:
fix https://github.com/astronomer/issues/issues/615
by creating whitelist and waiting for patch in buster image;

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -220,13 +220,18 @@ commands:
       image_name:
         type: string
         default: $CIRCLE_PROJECT_REPONAME
+      whitelist:
+        type: string
+        default: .circleci/cve-whitelist.yaml
     steps:
+      - checkout
       - attach_workspace:
           at: /tmp/workspace
       - run:
           name: Move tarball to directory for scan
           command: mkdir /tmp/tarballs && mv /tmp/workspace/<< parameters.image_name >>.tar /tmp/tarballs/
       - clair-scanner/scan:
+          whitelist: $CIRCLE_WORKING_DIRECTORY/cve-whitelist.yaml
           docker_tar_dir: /tmp/tarballs
   airflow-image-test:
     description: Test an Airflow image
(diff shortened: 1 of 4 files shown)


andriisoldatenko:
@ashb @schnie PTAL

.circleci/config.yml
```diff
@@ -298,3 +284,159 @@ commands:
...
+                munged_image=$(echo "$image" | sed 's/\//_/g' | sed 's/:/_/g')
+                sanitised_image_filename="${munged_image}.json"
+                local ret=0
+                local docker_cmd=(docker exec -it "$CLAIR_SCANNER" clair-scanner \
+                    --ip "$scanner_ip" \
+                    --clair=http://"$clair_ip":6060 \
+                    -t "<< parameters.severity_threshold >>" \
+                    --report "/$sanitised_image_filename" \
+                    --log "/log.json" \
+                    --whitelist /whitelist.yml \
```
andriisoldatenko: @sjmiller609 patched this line ^

cve-whitelist.yaml
```diff
@@ -0,0 +1,3 @@
+generalwhitelist:
+  CVE-2019-5482: curl
```
ashb: You've replaced curl with wget in the script -- do we still need this then?

andriisoldatenko: yes, we need `CVE-2019-5482: curl` to whitelist `git` dependency `libcurl3-gnutls`

ashb: If we're ignoring this then then we may as well stick with curl to install instead of wget?

.circleci/config.yml
ashb: One of the parameters to the orb is `whitelist`, and it looks like it (tries) to set the `WHITELIST` environment variable. At least acording to the source on https://circleci.com/orbs/registry/orb/ovotech/clair-scanner#orb-source it should have this line 94:

```
                    --log "/log.json" ${WHITELIST:+"-x"}
```

Which, if it worked would mean we don't have to build our own orb?

cve-whitelist.yaml
andriisoldatenko: yes

.circleci/config.yml
andriisoldatenko: https://github.com/ovotech/circleci-orbs/issues/89

cve-whitelist.yaml
andriisoldatenko: fixed.

PR #18, opened 2019-12-12
[astronomer/issues#650] Create Airflow 1.10.6 Image

kaxil:

#### Checklist
- [x] No new image, or image added to .circleci/config.yml jobs and workflow


kaxil:
> Guess this works, but I'm not hugely happy with the ammount of duplication we've got between the branches -- both in terms of the binary package files (are they re-compiled or copied across. They should be identical so we can just `git cp` them rather than rebuilding if that's what you did) and the Dockerfiles too.
> 
> Hmmm.

I just literally copy-pasted the **1.10.5** folder and changed the versions everywhere from **1.10.5-1** and **1.10.5-2** to **1.10.6-1**

kaxil:
@ashb I also agree that there is a good amount of duplication between these folders. Also, I haven't yet figured out a way to build 1.10.6 image using CircleCI without duplicating the entire Job in https://github.com/astronomer/ap-airflow/blob/master/.circleci/config.yml

ashb:
> I just literally copy-pasted the **1.10.5** folder and changed the versions everywhere from **1.10.5-1** and **1.10.5-2** to **1.10.6-1**

Yup, and that's probably okay for now if we want to get a 1.10.6 image out, and we can work out a way of reducing duplication in the future if we want to.

1.10.6/alpine3.10/Dockerfile
```diff
@@ -0,0 +1,127 @@
...
+	&& adduser -S -G ${ASTRONOMER_GROUP} ${ASTRONOMER_USER}
+
+COPY apks/keys/humans@astronomer.io.rsa.pub /etc/apk/keys
+
+# Force pip to install these specific versions when ever it installs a module
+ENV PIP_CONSTRAINT=/usr/local/share/astronomer-pip-constraints.txt
+COPY include/pip-constraints.txt /usr/local/share/astronomer-pip-constraints.txt
+
+# Install packages
+RUN echo https://github.com/astronomer/ap-airflow/raw/${REPO_BRANCH}/1.10.6/alpine3.10/apks/repo >> /etc/apk/repositories \
```
sjmiller609: It it possible to use a COPY line so it doesn't "reach into the past"?

kaxil:
Porting the question to this PR:

We seem to be missing `1.10.5/rhel7/onbuild/Dockerfile`. There is no **onbuild** directory in `1.10.5/rhel7` directory.

Is this intentional @ashb @schnie ?

The other test failure is the following:

```
ERROR: https://github.com/astronomer/ap-airflow/raw/master/1.10.6/alpine3.10/apks/repo: No such file or directory
```

that is because this PR hasn't been merged yet.

1.10.6/alpine3.10/Dockerfile
sjmiller609: I tried in my PR, but it didn't work for me.

sjmiller609:
> > I just literally copy-pasted the **1.10.5** folder and changed the versions everywhere from **1.10.5-1** and **1.10.5-2** to **1.10.6-1**
> 
> Yup, and that's probably okay for now if we want to get a 1.10.6 image out, and we can work out a way of reducing duplication in the future if we want to.

What about making a directory 'packages' in the git root, and moving 'apks' into that, then symlinking with a relative path?

ashb:
> Porting the question to this PR:
> 
> We seem to be missing `1.10.5/rhel7/onbuild/Dockerfile`. There is no **onbuild** directory in `1.10.5/rhel7` directory.
> 
> Is this intentional @ashb @schnie ?

No, probably not?

> 
> The other test failure is the following:
> 
> ```
> ERROR: https://github.com/astronomer/ap-airflow/raw/master/1.10.6/alpine3.10/apks/repo: No such file or directory
> ```
> 
> that is because this PR hasn't been merged yet.

There's a `ARG REPO_BRANCH=master` that specifies what branch to set request from. In Drone we used to set that based on what PR was building, but I don't think that got ported to our CircleCI config

ashb:
> > > I just literally copy-pasted the **1.10.5** folder and changed the versions everywhere from **1.10.5-1** and **1.10.5-2** to **1.10.6-1**
> > 
> > 
> > Yup, and that's probably okay for now if we want to get a 1.10.6 image out, and we can work out a way of reducing duplication in the future if we want to.
> 
> What about making a directory 'packages' in the git root, and moving 'apks' into that, then symlinking with a relative path?

We can probably do it without a symlink, as we fetch it via HTTP request, so we can just update the URL in the Dockerfiles. Lets do:

```
RUN echo https://github.com/astronomer/ap-airflow/raw/${REPO_BRANCH}/alpine_3.10_apks/repo >> /etc/apk/repositories \
```

kaxil:
> There's a `ARG REPO_BRANCH=master` that specifies what branch to set request from. In Drone we used to set that based on what PR was building, but I don't think that got ported to our CircleCI config

So somewhere we should set REPO_BRANCH=`CIRCLE_BRANCH`

kaxil:
WDYT of https://github.com/astronomer/ap-airflow/commit/4e6ab37383ad3896f31d72e8178c12cb30523324

> > > > I just literally copy-pasted the **1.10.5** folder and changed the versions everywhere from **1.10.5-1** and **1.10.5-2** to **1.10.6-1**
> > > 
> > > 
> > > Yup, and that's probably okay for now if we want to get a 1.10.6 image out, and we can work out a way of reducing duplication in the future if we want to.
> > 
> > 
> > What about making a directory 'packages' in the git root, and moving 'apks' into that, then symlinking with a relative path?
> 
> We can probably do it without a symlink, as we fetch it via HTTP request, so we can just update the URL in the Dockerfiles. Lets do:
> 
> ```
> RUN echo https://github.com/astronomer/ap-airflow/raw/${REPO_BRANCH}/alpine_3.10_apks/repo >> /etc/apk/repositories \
> ```

1.10.6/alpine3.10/Dockerfile
sjmiller609: discussed in PM

kaxil:
Had a call with @schnie and we decided that:

- We don't want to build RHEL image so should remove it from CircleCI config
- Pass the build branch at runtime to REPO_BRANCH argument
- We would handle the APK duplication at the later stage. I tried with the solution Ash suggested but `COPY apks/keys/humans@astronomer.io.rsa.pub /etc/apk/keys` command fails without it.

sjmiller609:
> Had a call with @schnie and we decided that:
> 
> * We don't want to build RHEL image so should remove it from CircleCI config
> * Pass the build branch at runtime to REPO_BRANCH argument
> * We would handle the APK duplication at the later stage. I tried with the solution Ash suggested but `COPY apks/keys/humans@astronomer.io.rsa.pub /etc/apk/keys` command fails without it.

How is the RHEL image used?

kaxil:
> > Had a call with @schnie and we decided that:
> > 
> > * We don't want to build RHEL image so should remove it from CircleCI config
> > * Pass the build branch at runtime to REPO_BRANCH argument
> > * We would handle the APK duplication at the later stage. I tried with the solution Ash suggested but `COPY apks/keys/humans@astronomer.io.rsa.pub /etc/apk/keys` command fails without it.
> 
> How is the RHEL image used?

It is not used anywhere. A customer committed the code so we have stored it if we might want to use it in the future.

sjmiller609:
Ok I see. I recommend we delete it until it is used. We are not at risk to lose it since it's in the commit history. Part of our security policy states that we will delete unused code.

schnie:
We need a home for our RHEL images so it may as well be here with our others. It is being used by Marriott, we just don't need to publish it to public repos just yet. So I'd say keep it in here for now just not auto-build and push yet.

sjmiller609:
I'm just not really a fan of commenting out tests. These images will have security vulnerabilities if we are not scanning them and remediating issues.

sjmiller609:
I made another PR to this branch showing how we can still accomplish testing and scanning for rhel7 but not worry about airflow 1.10.6 for rhel7 https://github.com/astronomer/ap-airflow/pull/21

sjmiller609:
there are 7 high-severity vulnerabilities in the rhel image https://987-221298116-gh.circle-artifacts.com/0/clair-reports/ap-airflow_1.10.5-rhel7.json

kaxil:
> there are 7 high-severity vulnerabilities in the rhel image https://987-221298116-gh.circle-artifacts.com/0/clair-reports/ap-airflow_1.10.5-rhel7.json

I think we should solve the vulnerabilities for RHEL as a separate PR, it isn't actually a part of this Issue unless it is straightforward to fix them. I think we should not delay releasing 1.10.6 alpine and buster images because of it.

sjmiller609:
That makes sense to me. We could merge this and allow the circle CI badge to correctly show failing build.

ashb:
>     * We would handle the APK duplication at the later stage. I tried with the solution Ash suggested but `COPY apks/keys/humans@astronomer.io.rsa.pub /etc/apk/keys` command fails without it.

Oh of course. We would need to manually `cp` that file in to somewhere under each folder for docker's `COPY` to be able to see it.

PR #19, opened 2019-12-12
only check if AIRFLOW__CORE__SQL_ALCHEMY_CONN is set than loop and wait

andriisoldatenko:
fix https://github.com/astronomer/issues/issues/652#issuecomment-564823312


ashb:
(@kaxil This is why I'm slightly wary of the copy-and-paste approach for a 1.10.6 image)

schnie:
@ashb came here to say the same thing haha. Same kind of thing happened with the RHEL image.

PR #20, opened 2019-12-13
DRY CI

sjmiller609:

Single point of control over change: new distribution or airflow version

You provide just a list of each and the rest of the CI is automatic with no change required.

- Made the circle config as DRY as possible, but it's not possible yet to be completely DRY, so I added a small script and Jinja2 template for the config file
- ~~Renamed alpine3.10 directories to 'alpine' to conform to convention of other distros in this repo (and took care of all places where we needed to swap that out)~~
- Modified the pull request template to document the procedure for adding a new distribution or airflow version
- re-ordered the function in the test script so the tests are at the top. This makes it easier to read, and more likely contributors will write tests

#### Checklist
- [x] No new image, or image added to .circleci/config.yml jobs and workflow


schnie:
@sjmiller609 why did we rename alpine? This is following the standard that the docker library images follow? What if we do want to support the next version of alpine as well?

sjmiller609:
> @sjmiller609 why did we rename alpine? This is following the standard that the docker library images follow? What if we do want to support the next version of alpine as well?
> 
> This actually breaks conformity because `buster` is short for `debian-buster`, `alpine3.10` encapsulates both the distro and the version.

I want to be able to configure just a list of distros and airflow versions. We can do 'alpine3.10', but we will not release ap-airflow:1.10.5-alpine, just ap-airflow:1.10.5-alpine3.10. So either works for me, just needs to not have an exception compared to the other distros (no special other tag)

schnie:
Ok cool, I'd say lets roll back to `alpine3.10` for the directory / disto and just stop publishing the `-alpine` tag.

schnie:
@sjmiller609 are we publishing the `-$BUILD_NUMBER` version of the tag like we have now too?

sjmiller609:
what's going on with this line? https://github.com/astronomer/ap-airflow/blob/d960717337b1413052af4b271dd69a24ce5cfda9/1.10.5/alpine3.10/Dockerfile#L52 seems weird to grab from somewhere else and also it's not working for me. I think maybe I'm just missing something obvious here

sjmiller609:
there is an assumption baked into this that all airflow versions support all the same distributions. is that correct?

sjmiller609:
> @sjmiller609 are we publishing the `-$BUILD_NUMBER` version of the tag like we have now too?

no, do you want it?

sjmiller609:
The only reason this pipeline is failing is because the build references something that doesn't exist. same problem exists in the upstream of this. We ought to use COPY instead of RUN if we can in this case, but I think I'm missing something that's happening because I don't see that file in the repository.

kaxil:
Yes, we seem to be missing `1.10.5/rhel7/onbuild/Dockerfile`. There is no **onbuild** directory in `1.10.5/rhel7` directory.

Is this intentional @ashb @schnie ?

sjmiller609:
If we are happy with the process, do you want to merge this into your branch, then we can look at the build errors?

kaxil:
> If we are happy with the process, do you want to merge this into your branch, then we can look at the build errors?

Yes definitely, good work with using the templating engine (Jinja) for this.

PR #24, opened 2019-12-19
Accept/Ignore Kernel vulnerabilities in RHEL image

ashb:
Docker images don't run their own kernel, so these aren't an issue for
us

#### Checklist


ashb:
This doesn't seem to be the right format. Next to no docs on the exact format of `whitelist.yml` :(

Any idea @andriisoldatenko

PR #27, opened 2019-12-20
Release 1.10.5-3

kaxil:


1.10.5/buster/Dockerfile
```diff
@@ -106,7 +106,7 @@ RUN apt-get update \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/*
 
-ARG VERSION="1.10.5-1"
```
kaxil: @schnie Why is this **1.10.5-1** whereas the alpine one is **1.10.5-2**

schnie: Must have been committed while we were transitioning or something. Good eye! Thanks for fixing!

PR #36, opened 2020-02-04
Add test to confirm Redis image version

sjmiller609:

@ashb for this comment: https://astronomerteam.slack.com/archives/CLRKTKHGU/p1580586195005300

#### Checklist
n/a

### DIFF
--- a/.circleci/test-airflow-image.py
+++ b/.circleci/test-airflow-image.py
@@ -62,7 +62,7 @@ def test_redis_version(webserver):
     except KeyError:
         raise Exception("redis pip module is not installed")
     version = redis_module['version']
-    assert semantic_version(version) != semantic_version('3.4.0'),
+    assert semantic_version(version) != semantic_version('3.4.0'), \
            "redis module must not be 3.4.0"
 
 @pytest.fixture(scope='session')
--- a/.circleci/test-airflow-image.py
+++ b/.circleci/test-airflow-image.py
@@ -55,15 +55,15 @@ def test_elasticsearch_version(webserver):
 
 
 def test_redis_version(webserver):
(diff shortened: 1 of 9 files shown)


ashb:
Yes, but 3.4.1 is released that fixes the problem :)

sjmiller609:
> Yes, but 3.4.1 is released that fixes the problem :)

@ashb do we want to check >= 3.4.1?

ashb:
Yes, sounds good

I'm on mobile so can't check failures easily - are they falling this new check?

sjmiller609:
looks like an unexpected error with CI "no space left on device", i'll just re-run

.circleci/test-airflow-image.py
```diff
@@ -54,6 +54,17 @@ def test_elasticsearch_version(webserver):
         "elasticsearch module must be version 5.5.3 or greater"
 
 
+def test_redis_version(webserver):
+    """ Astronomer requires a version of redis greater than 3.4.1
```
ashb:
```suggestion
    """ Astronomer requires a version of redis greater than or equal to 3.4.1
```

sjmiller609:
Looks like some images are 3.3.11, so I made it check >= 3.4.1 or == 3.3.11

ashb:
Ah yeah, ones that haven't been rebuilt since will fail I guess.

Do you think its worth adding a

```
redis!=3.4.0
```

to all the pip-constraint.txt files?

sjmiller609:
Ok I switched up the test logic to test that it's NOT 3.4.0 and also I did the pip-constraints change as requested. rhel is failing because of a security issue.

sjmiller609:
@schnie this one is low risk and fixes an active issue

PR #37, opened 2020-02-06
Ensure tini present, ensure werkzeug version, fix tests

sjmiller609:

I am baking an assumption on the location of the entrypoint into Astronomer. This test serves to ensure the expected functionality is retained.

There is an [issue](https://github.com/astronomer/issues/issues/620) that the airflow scheduler overwrites the entrypoint in KPO and k8s executor workers. To solve that issue, I am modifying the pod mutation hook in airflow to make use of our entrypoint, which is invoked with 'tini'

#### Checklist

N/A

### DIFF
--- a/.circleci/test-airflow-image.py
+++ b/.circleci/test-airflow-image.py
@@ -94,6 +94,7 @@ def webserver(request):
     https://testinfra.readthedocs.io/en/latest/examples.html#test-docker-images
     """
     docker_id_db = start_postgres()
+    wait_for_container(docker_id_db)
     db_connection_string = f"postgres://postgres:notsecretpassword@{get_ip_from_id(docker_id_db)}:5432"
     docker_id = subprocess.check_output(
         ['docker', 'run', '--rm',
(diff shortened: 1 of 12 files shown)


schnie:
@sjmiller609 why are these failing?

sjmiller609:
I think I'm finding this issue in a few of the images https://github.com/pallets/werkzeug/issues/1616

sjmiller609:
I think werkzeug 1.16.0 https://werkzeug.palletsprojects.com/en/0.16.x/changes/#version-0-16-0

sjmiller609:
@ashb Greg mentioned you know about this one, any tips?

sjmiller609:
got it https://github.com/apache/airflow/commit/4b25cb9d08565502172cb847c79d81559775d504

thanks for easy to follow paper trail ash :)

sjmiller609:
it was failing for some reason I was unable to reproduce locally. that was because it was a race condition between airflow shutting down and the tests finishing. Airflow is failing out with this error:
```
[2020-02-07 23:23:09,713] {dagbag.py:243} ERROR - Failed to import: /usr/local/lib/python3.7/site-packages/airflow/example_dags/example_subdag_operator.py
```

sjmiller609:
I believe the answer is to initdb first.

PR #39, opened 2020-02-10
It's "CEA", not "Certified Airflow"

ryw:

#### Checklist


schnie:
Good catch!

PR #40, opened 2020-02-21
Papermill 2.0 includes PyArrow, which is hard to install on Alpine

ashb:
To avoid this, we just restrict papermill to <2 only for Alpine images.

#### Checklist


ashb:
This is a bit of a hack I guess, but Papermill 2.0 is only 2.0 because they dropped support for Py3.5, so other than pulling in pyarrow it's not a problem to be on a newer version on Debian.

pyarrow is failing all sorts of cmake/compilation issues on Alpine that I just don't want to dig in to now.

ashb:
https://gist.github.com/bskaggs/fc3c8d0d553be54e2645616236fdc8c6 will be useful in the future (unless we just drop Alpine images)

PR #41, opened 2020-02-24
Install apache-airflow from our pre-packaged wheel

ashb:

This saves us having to install then remove nodejs. It also makes the docker image "the same" as the python install of CEA

#### Checklist

### DIFF
--- a/1.10.7/alpine3.10/Dockerfile
+++ b/1.10.7/alpine3.10/Dockerfile
@@ -23,16 +23,13 @@ LABEL io.astronomer.docker.component="airflow"
 LABEL io.astronomer.docker.airflow.version="1.10.7"
 
 ARG ORG="astronomer"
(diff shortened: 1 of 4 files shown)


schnie:
LGTM

ashb:
(Let's wait for some testing before merging this)

ashb:
@andrewhharmon Any further feedback? We should merge this soon if we're going to.

andrewhharmon:
I was able to do a pip install over OSS Airflow 1.10.7 with no issues. I also tried a pip install over OSS Airflow 1.10.9 with no issues. Se may be lucky that there are no breaking db migrations between 1.10.7 and 1.10.9.  So LGTM

Commands I used. 

pip3 install --extra-index-url=https://pip.astronomer.io/simple/ 'apache-airflow>=1!1.10.7'
airlfow upgradedb

PR #48, opened 2020-03-13
Fix airflow worker grace period on Buster-based images

ashb:

The problem was that PID1 (the tini process) was run as what ever the "default" user was, which when run on our platform is currently hard-coded to UID=100. We'd put in steps to switch to the other user to run airflow (via sudo) but unfortunately this meant when we tried to "gracefully shut down the celery worker" it would send the signal to tini, which would try to forward it on to sudo, but this would fail with:

    [FATAL tini (1)] Unexpected error when forwarding signal: 'Operation not permitted'

So the fix foe this is to make sure what ever is PID1 is running as the right user.

The way we achieve this is:

1. Switch to `gosu` (it was already installed) as this doesn't leave a "sudo" process in the middle. If we kept sudo we would have sudo as PID1 which would handle signals properly (PID1 is special)
2. gosu doesn't have a config file -- you need to be root to be able to run it. To work around this we make it a sticky/SUID executable, so that anyone who can run it (via normal file permissions) will run it "as root"
3. To "limit the blast radius" of gosu, we only let root and members of the new "gosuers" group be able to execute the binary.

The end result of this: `tini` is PID1, and running as the "correct"
astro user, even the container was launched as UID 100, and signals
correctly propagate where needed

Fixes astronomer/issues#718

/cc @kaxil You'll need to update this in your PR that adds 1.10.9 too. (We really need to sort out duplication of all these dockerfiles/scripts!)

### DIFF
--- a/1.10.10.dev/buster/Dockerfile
+++ b/1.10.10.dev/buster/Dockerfile
@@ -142,14 +142,17 @@ COPY include/pip-constraints.txt /usr/local/share/astronomer-pip-constraints.txt
 RUN install --directory --owner="${ASTRONOMER_USER}" "${AIRFLOW_HOME}" \
     && install --directory --owner="${ASTRONOMER_USER}" "${AIRFLOW_HOME}/logs"
 
-COPY include/sudoers /etc/sudoers.d/astro
-RUN visudo --check --file=/etc/sudoers.d/astro
-
 # Copy entrypoint to root
 COPY include/entrypoint /
 
 # Copy "cron" scripts
 COPY include/clean-airflow-logs /usr/local/bin/clean-airflow-logs
+
+# Set it up so that _apt/UID 100 can `gosu`, but no other users can
+RUN groupadd gosuers \
+    && usermod --append --groups gosuers _apt \
+    && chgrp gosuers /usr/sbin/gosu \
+    && chmod u+s,g+sx,o-rx /usr/sbin/gosu
 
 # Though this is set here we currently override this in the helm template, so
 # this _might_ not have any effect once deployed. The /entrypoint script copes
@@ -159,5 +162,5 @@ USER ${ASTRONOMER_USER}
 # Switch to AIRFLOW_HOME
 WORKDIR ${AIRFLOW_HOME}
 
-ENTRYPOINT ["tini", "--", "/entrypoint"]
+ENTRYPOINT ["/entrypoint"]
 CMD ["airflow", "--help"]
--- a/1.10.10.dev/buster/include/entrypoint
+++ b/1.10.10.dev/buster/include/entrypoint
@@ -1,10 +1,20 @@
 #!/usr/bin/env bash
 set -e
 
-# Sudo _explicitly_ filters out PYTHONPATH, even when env_reset is disabled. but we want it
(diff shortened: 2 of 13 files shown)


ashb:
Sorry for the hacks I have committed here @jpweber - see my latest comments in https://github.com/astronomer/issues/issues/718 for why I'm doing such a monstrous thing.

ashb:
Hmmm -CrashLoopBackoff on the scheduler pod (that I didn't get locally)

schnie:
@ashb seems like if Houston was aware of the version/distro we could avoid all this and just pass in the right `uid` right? Should this be something we aim to cleanup with that project?

https://github.com/astronomer/airflow-chart/blob/master/values.yaml#L5-L7

ashb:
@schnie Yeah we should fix this properly on the houston side.

And I think since my "fetch docker metadata on push" change we can now tell via the `LABEL io.astronomer.docker.distro="debian"`.

We could also look at `LABEL io.astronomer.docker.uid="${ASTRONOMER_UID}"` -- but we probably don't want to "blindly" trust that, or at least assert it is `== 100 or >= 1000` say.

This is just a short-workaround to fix the bug for now

ashb:
This change has (now) removed `sudo` rights from the `astro` user. That may cause problems for downstream users who wish to do `sudo apt-get` commands etc. There are better ways of doing it already (our onbuild image for instance), but do we want to continue to support the "bad" way?

schnie:
@andrewhharmon holler when you confim this and we'll merge it on in.

schnie:
@vishwas-astro we'll be able to back this code out once we solve the "version/distro-awareness" thing we chatted about. It won't hurt though, just gets ignored.

schnie:
@ashb looks like we got conflicts now that we consolidated the `onbuild` stuff.

ashb:
@schnie Conflicts resolved.

ashb:
> Please be aware that the entrypoint is assumed here https://github.com/astronomer/astronomer/blob/e9b3271b040e816ef1366341ee9cda6cd2c4df70/charts/astronomer/templates/houston/houston-configmap.yaml#L156

That is only for pods launched by Kube executor/pod operator, which wasn't the subject of this PR, so I think this isn't problem

ashb:
Oh I've removed the ability for the `astro` user to run sudo, that's probably going to be annoying isn't it @andrewhharmon

sjmiller609:
@schnie only you are authorized to merge this one.

ashb:
Nope, Kube executor is fine, cos there is no "warm shutdown but stay up until done" in the executor, just in the celery worker

ashb:
I need to update the 1.10.10 Dockerfile too

kaxil:
> I need to update the 1.10.10 Dockerfile too

Weird, did you force-push, I had applied changes to 1.10.10 earlier

kaxil:
Check this commit: 37e7f66

ashb:
> Oh I've removed the ability for the `astro` user to run sudo, that's probably going to be annoying isn't it @andrewhharmon

Just noticed we don't have sudoers file in Alpine, so I'm leaving it out/removing it from our buster image. (There is always `USER` dockerfile cmd if needed.)

PR #52, opened 2020-03-23
Replace installing apache-airflow with astronomer_certified for 1.10.7

kaxil:
We have published a wheel package for AC Airflow 1.10.7: 
https://pip.astronomer.io/simple/astronomer-certified/astronomer_certified-1.10.7.post7-py2.py3-none-any.whl

This PR replaces installing apache-airflow package with astronomer_certified for 1.10.7


schnie:
Are we going to leave our prior releases alone or also push a package for them?

kaxil:
> Are we going to leave our prior releases alone or also push a package for them?

AFAIK 1.10.7 would be the first supported version so I think we would leave prior releases as it is.

PR #56, opened 2020-03-30
Dont use cache when building docker images

kaxil:
Because CircleCI was using cache it was using the same whl package to build the dev docker images

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -382,7 +382,7 @@ workflows:
           airflow_version: 1.10.10.dev
           distribution_name: alpine3.10
           docker_repo: astronomerio/ap-airflow
-          extra_args: "--build-arg VERSION='=='$(curl https://pip.astronomer.io/simple/astronomer-certified/latest-1.10.10.build)"
+          extra_args: "--build-arg VERSION=$(curl https://pip.astronomer.io/simple/astronomer-certified/latest-1.10.10.build)"
       - scan:
           name: scan-1.10.10.dev-alpine3.10-onbuild
           airflow_version: 1.10.10.dev
@@ -435,7 +435,7 @@ workflows:
           airflow_version: 1.10.10.dev
           distribution_name: buster
           docker_repo: astronomerio/ap-airflow
-          extra_args: "--build-arg VERSION='=='$(curl https://pip.astronomer.io/simple/astronomer-certified/latest-1.10.10.build)"
+          extra_args: "--build-arg VERSION=$(curl https://pip.astronomer.io/simple/astronomer-certified/latest-1.10.10.build)"
       - scan:
           name: scan-1.10.10.dev-buster-onbuild
           airflow_version: 1.10.10.dev
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -497,6 +497,7 @@ workflows:
           airflow_version: 1.10.10.dev
           distribution_name: alpine3.10
           docker_repo: "astronomerio/ap-airflow"
+          extra_args: "--build-arg VERSION="$(curl https://pip.astronomer.io/simple/apache-airflow/latest-1.10.10.build)""
       - scan:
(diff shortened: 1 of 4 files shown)


ashb:
Not the right fix -- by not using the cache each image will share no layers with the previous one, so images will be bigger and take longer to build.

kaxil:
> Not the right fix -- by not using the cache each image will share no layers with the previous one, so images will be bigger and take longer to build.

Updated

.circleci/config.yml
```diff
@@ -497,6 +499,7 @@ workflows:
           airflow_version: 1.10.10.dev
           distribution_name: alpine3.10
           docker_repo: "astronomerio/ap-airflow"
+          extra_args: "--build-arg VERSION='=='$(curl https://pip.astronomer.io/simple/astronomer-certified/latest-1.10.10.build)"
```
ashb: You edited this manually, and didn't update the templates/the rest to match.

.circleci/config.yml.j2
```diff
@@ -11,6 +11,9 @@ workflows:
           airflow_version: {{ airflow_version }}
           distribution_name: {{ distribution }}
           docker_repo: {{ docker_repo }}
+          {%- if "dev" in airflow_version %}
+          extra_args: "--build-arg VERSION=$(curl https://pip.astronomer.io/simple/astronomer-certified/latest-{{ airflow_version | replace('.dev', '') }}.build)"
```
ashb:
```suggestion
          extra_args: "--build-arg VERSION===$(curl https://pip.astronomer.io/simple/astronomer-certified/latest-{{ airflow_version | replace('.dev', '') }}.build)"
```

(no need to quote `=` as it's not a special shell character.

.circleci/config.yml
kaxil: Nope, didn't edit manually

kaxil: I added conditions in the template file to only use this for DEV version

1.10.10.dev/buster/Dockerfile
```diff
@@ -103,17 +103,17 @@ RUN apt-get update \
...
+ARG AIRFLOW_MODULE="astronomer_certified[${SUBMODULES}]==$VERSION"
 
 # Make pip look at our pip repo too, and force it to install these specific
 # versions when ever it installs a module.
 COPY include/pip.conf /etc/pip.conf
 COPY include/pip-constraints.txt /usr/local/share/astronomer-pip-constraints.txt
 
 # Pip install airflow and astro security manager
-RUN pip install "${AIRFLOW_MODULE}" --pre \
+RUN pip install --no-cache-dir "${AIRFLOW_MODULE}" --pre \
```
ashb: Don't need this change anymore

1.10.10.dev/buster/Dockerfile
```diff
@@ -103,17 +103,17 @@ RUN apt-get update \
...
   && pip install "https://github.com/astronomer/astronomer-airflow-scripts/releases/download/v0.0.4/astronomer_airflow_scripts-0.0.4-py3-none-any.whl" \
 	&& pip install --no-cache-dir "astronomer-fab-security-manager~=1.2, >=1.2.2"
```
ashb:
```suggestion
	&& pip install "astronomer-fab-security-manager~=1.2, >=1.2.2"
```

since we set that globally in via env in the container

kaxil: Updated

1.10.10.dev/buster/Dockerfile
kaxil: Updated

.circleci/config.yml.j2
kaxil: I moved '==' inside the Dockerfile

kaxil: Actually moved that logic inside Dockerfile

PR #58, opened 2020-04-07
[astronomer/issues#1042] Add appropriate '-number' tags to Docker images

kaxil:
Add the -number to the end of our docker tags, for example `1.10.7-7` for the `astro.7`

Closes https://github.com/astronomer/issues/issues/1042

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -38,7 +38,7 @@ workflows:
           name: publish-1.10.5-alpine3.10
           docker_repo: astronomerinc/ap-airflow
           tag: "1.10.5-alpine3.10"
-          extra_tags: "1.10.5-alpine3.10-${CIRCLE_BUILD_NUM}"
+          extra_tags: "1.10.5-alpine3.10-${CIRCLE_BUILD_NUM},1.10.5-6-alpine3.10"
           requires:
             - scan-1.10.5-alpine3.10-onbuild
             - scan-trivy-1.10.5-alpine3.10-onbuild
@@ -50,7 +50,7 @@ workflows:
           name: publish-1.10.5-alpine3.10-onbuild
           docker_repo: astronomerinc/ap-airflow
           tag: "1.10.5-alpine3.10-onbuild"
-          extra_tags: "1.10.5-alpine3.10-onbuild-${CIRCLE_BUILD_NUM}"
+          extra_tags: "1.10.5-alpine3.10-onbuild-${CIRCLE_BUILD_NUM},1.10.5-6-alpine3.10-onbuild"
           requires:
             - scan-1.10.5-alpine3.10-onbuild
             - scan-trivy-1.10.5-alpine3.10-onbuild
@@ -90,7 +90,7 @@ workflows:
           name: publish-1.10.5-buster
           docker_repo: astronomerinc/ap-airflow
           tag: "1.10.5-buster"
-          extra_tags: "1.10.5-buster-${CIRCLE_BUILD_NUM}"
+          extra_tags: "1.10.5-buster-${CIRCLE_BUILD_NUM},1.10.5-6-buster"
           requires:
             - scan-1.10.5-buster-onbuild
             - scan-trivy-1.10.5-buster-onbuild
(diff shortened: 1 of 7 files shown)


.circleci/config.yml
```diff
@@ -328,6 +334,7 @@ workflows:
           airflow_version: 1.10.7
           distribution_name: buster
           docker_repo: astronomerinc/ap-airflow
+          extra_args: "-t astronomerinc/ap-airflow:1.10.7-7-buster"
```
ashb, line 337: Does this actually push up the extra tag too?

ashb: Given there's 

```
      - publish:
          name: publish-1.10.7-buster
```

I don't think this change alone is enough?

ashb: And we only do

```
            docker save -o saved-images/<< parameters.image_name >>.tar << parameters.image_name >>
```

so only one image will get loaded back in to bet pushed.

kaxil: Yeah, it won't.  Fixing it

kaxil: Updated. I think this approach is better. And it should work because of https://github.com/astronomer/ap-airflow/blob/2579903fa083808e4950fe7ab2f5b4b0c44a5275/.circleci/config.yml#L837-L841

.circleci/generate_circleci_config.py
```diff
@@ -9,9 +9,9 @@
 from jinja2 import Environment, FileSystemLoader
 
 IMAGE_MAP = collections.OrderedDict([
-    ("1.10.5", ["alpine3.10", "buster", "rhel7"]),
-    ("1.10.6", ["alpine3.10", "buster"]),
-    ("1.10.7", ["alpine3.10", "buster"]),
+    ("1.10.5-6", ["alpine3.10", "buster", "rhel7"]),
```
ashb, line 12: Hmmm hmmm hmmm.

I wonder if we could extract this info out of the image, just so we don't have to (remember, and fail) to update this info in two places.

ashb: Other than the fact that we don't currently set the right version:


```
$ docker image inspect astronomerinc/ap-airflow:1.10.7-alpine3.10-onbuild-4208 -f '{{index .Config.Labels "io.astronomer.docker.airflow.version"}}'
1.10.7
```

kaxil: Yeah, I think it has been there since the start and I feel the intention was to show **Airflow Version**:

https://github.com/astronomer/ap-airflow/blob/10167d89afd233a9633496456d4c240e1da2d5d5/1.10.7/alpine3.10/Dockerfile#L23

https://github.com/astronomer/ap-airflow/blob/10167d89afd233a9633496456d4c240e1da2d5d5/1.10.7/alpine3.10/Dockerfile#L26

kaxil: Based on our discussion added a new label to specify AC Version

kaxil: Based on that commit, we will need to change the VERSION only in `generate_circleci_config.py` and it will change the VERSIONS in all Dockerfiles

1.10.7/buster/Dockerfile
```diff
@@ -21,12 +21,14 @@ LABEL maintainer="Astronomer <humans@astronomer.io>"
...
 ARG ASTRONOMER_USER="astro"
 ARG ASTRONOMER_UID="50000"
+ARG VERSION="1.10.7-7"
 
 LABEL io.astronomer.docker=true
 LABEL io.astronomer.docker.distro="debian"
 LABEL io.astronomer.docker.module="airflow"
 LABEL io.astronomer.docker.component="airflow"
 LABEL io.astronomer.docker.uid="${ASTRONOMER_UID}"
+LABEL io.astronomer.docker.ac.version="${VERSION}"
```
ashb: This one (and the arg) shouldn't moved up here - it will blow the docker layer cache ntoo early

kaxil: I had updated that in the PR, can you double check the PR please or the last fixup commit

.circleci/config.yml
```diff
@@ -834,6 +834,12 @@ commands:
             set -x
             docker push "$image"
             set +x
+            # Tag AC Version for non-dev images
+            if [[ "<< parameters.image_name >>" != *"dev"* ]]; then
+                ac_version=$(docker image inspect $image -f '{{index .Config.Labels "io.astronomer.docker.ac.version"}}')
```
ashb, line 839: This will just be `1.10.7-7`, but we also need to include the distribtion in here too.

kaxil: fixed in https://github.com/astronomer/ap-airflow/pull/58/commits/9a2e658b594ad7ec58778f36c99b1925f708b8fb

PR #59, opened 2020-04-08
Fix entrypoints to correctly wait for DB/broker when no password

ashb:
If you specify the celery broker URL as `redis://host:6379/0` it would
fail to correctly extract the host/port and would end up in a loop never actually run.

The problem was the awk pipeline required a username/password. So
instead replace it with a regex. Hopefully now we don't have two problems

Running these image with `AIRFLOW__CELERY__BROKER_URL=redis://redis:6379/0` wasn't possible.

(I'm starting to hate the amount of duplication we have with all these entrypoints being _almost_ identical. All the alpine ones should be the same, and all the buster ones, but there are differences from alpine to buster -- alpine has an `airflow sync_perm` that buster doesn't.)

#### Checklist

### DIFF
--- a/1.10.10.dev/alpine3.10/include/entrypoint
+++ b/1.10.10.dev/alpine3.10/include/entrypoint
@@ -10,11 +10,14 @@ fi
 # Airflow subcommand
 CMD=$2
 
+url_parse_regex="[^:]+://([^@/]*@)?([^/:]*):?([0-9]*)/?"
+
 # Wait for postgres then init the db
 if [[ -n $AIRFLOW__CORE__SQL_ALCHEMY_CONN  ]]; then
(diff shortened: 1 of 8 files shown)


schnie:
Just to be sure - did we test this on our stuff? Just double checking before merging since this could cause some havoc if we unleash it.

ashb:
Haven't pushed this to a deployment (I tested the image locally with a few inputs), I should do that to double check and confirm, then I will merge

ashb:
```
Waiting for host: zodiacal-inertia-3389-pgbouncer.astronomer-zodiacal-inertia-3389.svc.cluster.local 6543
Waiting for host: zodiacal-inertia-3389-redis 6379
```

All good.

PR #66, opened 2020-04-22
Release Astro Certified 1.10.10-1

kaxil:
Issue: https://github.com/astronomer/issues/issues/1143
AC Tag: https://github.com/astronomer/airflow/releases/tag/v1.10.10%2Bastro.1

PR #68, opened 2020-04-23
Fix openssl CVE

kaxil:
Upgrade OpenSSL to fix the following CVE:

* https://cve.mitre.org/cgi-bin/cvename.cgi?name=2020-1967

### DIFF
--- a/1.10.5/alpine3.10/Dockerfile
+++ b/1.10.5/alpine3.10/Dockerfile
@@ -88,6 +88,8 @@ RUN echo https://github.com/astronomer/ap-airflow/raw/${REPO_BRANCH}/alpine-pack
 		py3-typed-ast \
 		python3 \
 		tini \
+	&& apk add --upgrade \
+		openssl \
 	&& update-ca-certificates \
 	&& cp /usr/share/zoneinfo/UTC /etc/localtime \
 	&& pip3 install --no-cache-dir --upgrade pip==19.3.1 \
--- a/1.10.6/alpine3.10/Dockerfile
+++ b/1.10.6/alpine3.10/Dockerfile
@@ -67,7 +67,6 @@ RUN echo https://github.com/astronomer/ap-airflow/raw/${REPO_BRANCH}/alpine-pack
 		python3-dev \
 		tzdata \
 	&& apk add --no-cache \
-		openssl \
 		bash \
 		ca-certificates \
 		cyrus-sasl \
--- a/1.10.6/alpine3.10/Dockerfile
+++ b/1.10.6/alpine3.10/Dockerfile
@@ -67,6 +67,7 @@ RUN echo https://github.com/astronomer/ap-airflow/raw/${REPO_BRANCH}/alpine-pack
 		python3-dev \
 		tzdata \
 	&& apk add --no-cache \
+		openssl \
 		bash \
 		ca-certificates \
 		cyrus-sasl \
@@ -88,6 +89,8 @@ RUN echo https://github.com/astronomer/ap-airflow/raw/${REPO_BRANCH}/alpine-pack
 		py3-typed-ast \
 		python3 \
 		tini \
+	&& apk add --upgrade \
+		openssl \
 	&& update-ca-certificates \
 	&& cp /usr/share/zoneinfo/UTC /etc/localtime \
 	&& pip3 install --no-cache-dir --upgrade pip==19.3.1 \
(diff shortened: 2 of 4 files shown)


1.10.5/alpine3.10/Dockerfile
```diff
@@ -88,6 +89,8 @@ RUN echo https://github.com/astronomer/ap-airflow/raw/${REPO_BRANCH}/alpine-pack
 		py3-typed-ast \
 		python3 \
 		tini \
+	&& apk add --upgrade \
+		openssl \
```
ashb, line 93: Why is this here and above?

kaxil: Removed from above.

kaxil:
> We need to update debian images too don't we?

The buster one already does `apt-get upgrade` - Wouldn't that update all the packages to the latest version? And Trivy is also only failing for Alpine images

ashb:
We probably don't want to blindly update alpine packages -- that could introduce a version change, where as for Debian (because of how the debian project choses to release packages) they never introduce new versions, only (security) fixes.

1.10.6/alpine3.10/Dockerfile
```diff
@@ -67,6 +67,7 @@ RUN echo https://github.com/astronomer/ap-airflow/raw/${REPO_BRANCH}/alpine-pack
 		python3-dev \
 		tzdata \
 	&& apk add --no-cache \
+		openssl \
```
ashb, line 70: Missed one.

ashb:
@kaxil Any idea why the 1.10.10 image didn't get an updated openssl, but the other ones did?

kaxil:
> @kaxil Any idea why the 1.10.10 image didn't get an updated openssl, but the other ones did?

The alpine one did, I am checking what happened with the buster image for 1.10.10.dev

kaxil:
> > @kaxil Any idea why the 1.10.10 image didn't get an updated openssl, but the other ones did?
> 
> The alpine one did, I am checking what happened with the buster image for 1.10.10.dev

It looks to me that for 1.10.10 buster CircleCI is using the cache image from a different build:

```
Found a cache from build 7439 at trivy-cache
Size: 15 MiB
Cached paths:
  * /tmp/workspace/trivy-cache

Downloading cache archive...
Validating cache...

Unarchiving cache...
```

and hence the issue.

ashb:
The trivy-cache is just for the info about vulnerabilities, not what is scanned.

kaxil:
> The trivy-cache is just for the info about vulnerabilities, not what is scanned.

Correct, my bad. The failure is unrelated to the changes in the PR as we haven't updated anything to the buster images. We updated the Alpine 1.10.10dev image that has passed the Trivy scanner.

Restarted tests

ashb:
K I see the problem:

```
Step 35/50 : FROM ${APT_DEPS_IMAGE} as main
 
 ---> c95a8a3cfff5
 
Step 36/50 : LABEL io.astronomer.docker.airflow.version "1.10.10.dev"
 
 ---> Using cache
 
 ---> 5f8c6eeb3be9
 
Step 37/50 : LABEL io.astronomer.docker.ac.version "${VERSION}"
 
 ---> Using cache
 
 ---> 741b59a0f17b
 
Step 38/50 : RUN apt-get update     && apt-get upgrade -y --no-install-recommends     && apt-get clean     && rm -rf /var/lib/apt/lists/*
 ```

Docker is re-using the layer cache, as nothing "before" that point has changed (the python package is installed in a separate afterwards because of the multilayer image.

So we need something between lines 123 and 128 that busts the docker cache, otherwise it will not update packages. (THis is why Airflow has, or had, the "BUILD_EPOCH" type variable.

kaxil:
> K I see the problem:
> 
> ```
> Step 35/50 : FROM ${APT_DEPS_IMAGE} as main
>  
>  ---> c95a8a3cfff5
>  
> Step 36/50 : LABEL io.astronomer.docker.airflow.version "1.10.10.dev"
>  
>  ---> Using cache
>  
>  ---> 5f8c6eeb3be9
>  
> Step 37/50 : LABEL io.astronomer.docker.ac.version "${VERSION}"
>  
>  ---> Using cache
>  
>  ---> 741b59a0f17b
>  
> Step 38/50 : RUN apt-get update     && apt-get upgrade -y --no-install-recommends     && apt-get clean     && rm -rf /var/lib/apt/lists/*
> ```
> 
> Docker is re-using the layer cache, as nothing "before" that point has changed (the python package is installed in a separate afterwards because of the multilayer image.
> 
> So we need something between lines 123 and 128 that busts the docker cache, otherwise it will not update packages. (THis is why Airflow has, or had, the "BUILD_EPOCH" type variable.

Good catch, should be resolved when we release new image in https://github.com/astronomer/ap-airflow/pull/66

PR #69, opened 2020-04-23
Upgrade sqlite3 packages for Alpine 3.10

ashb:
We don't need to explicitly update it as they aren't installed yet.

Building a new image I see this in the output:

```
(73/77) Installing sqlite-libs (3.30.1-r1)
```

so it is being used.


ashb:
@ryw Can you dig out details of what the vulns were please?

ashb:
I've messed up the signature on the APK file somehow.

ashb:
> This is a bit low level for me. But is there no way to update "gcc" without having to add the patches and those additional APK files?

No, sadly not. So Alpine haven't made any newer (or patched) versions of GCC available in Alpine 3.10 (3.11 is the latest) so if we want to mitigate the vuln then we need to compile a new version.

ashb:
I think CVE-2019-16168 is the sqlite one, I can't find out what the gcc one is.

ashb:
I need to remove gcc from this - it's not needed and is a big change

ryw:
that's fine, let's add note to SECURITY.md w/ the decision, reason, and future mitigation plan (i'd say we wait til we can use alpine 3.11, with airflow running on python 3.8)

ashb:
I have removed gcc -- the vulnerability only applies to compiling PowerPC, which is so not worth the trouble of carrying a gcc package oursleves.

PR #70, opened 2020-04-23
Add Changelog info to readme

petedejoy:

This PR removes the (previously empty) top-level changelog in favor of changelogs at the version level. It also adds some info to our readme on where to find relevant changelogs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,0 @@
-# Changelog
-
-All notable changes to this project will be documented in this file.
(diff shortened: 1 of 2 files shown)


paolaperaza:
Noice @petedejoy - should we add a bullet point for 1.10.6 too since there's a folder for it in the repo, and then update with 1.10.10 as well?

petedejoy:
Can def add 1.10.10. I thought about 1.10.6 but we don't have a changelog in there since we didn't cut an AC version with it.

paolaperaza:
Ah yes, didn't see that - makes sense to keep out 1.10.6 then

kaxil:
Converted to draft as merging any PR to ap-airflow will rebuild images and there is an open-issues regarding new image.

PR #76, opened 2020-05-04
Add test to avoid installing astronomer-airflow-version-check 1.0.0

kaxil:
Based on PR: https://github.com/astronomer/astronomer-airflow-version-check/pull/10 & GH Issue: https://github.com/astronomer/issues/issues/1240

The version check plugin isn't installed in AC < 1.10.7 hence I have only added constraint to AC >=1.10.7

This constraint is not **REQUIRED** as the dependency on AC version is =~1.0 so it should always take the latest available version >1 but this constraint can act as a safeguard when there is a cached 1.0.0 version of astronomer-airflow-version-check.

### DIFF
--- a/.circleci/bin/test-airflow-image.py
+++ b/.circleci/bin/test-airflow-image.py
@@ -102,7 +102,7 @@ def test_astronomer_airflow_check_version(webserver):
         print("astronomer-airflow-version-check pip module is not installed")
         return
     version = version_check_module['version']
-    assert semantic_version(version) > semantic_version('1.0.0'), \
+    assert semantic_version(version) >= semantic_version('1.0.1'), \
         "astronomer-airflow-version-check module must be greater than 1.0.0"
 
 
--- a/.circleci/bin/test-airflow-image.py
+++ b/.circleci/bin/test-airflow-image.py
@@ -25,17 +25,20 @@ def test_airflow_in_path(webserver):
     assert webserver.exists('airflow'), \
         "Expected 'airflow' to be in PATH"
 
+
 def test_tini_in_path(webserver):
     """ Ensure 'tini' is in PATH
     """
     assert webserver.exists('tini'), \
         "Expected 'tini' to be in PATH"
 
+
 def test_entrypoint(webserver):
     """ There should be a file '/entrypoint'
     """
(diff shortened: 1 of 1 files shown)


1.10.10/alpine3.10/include/pip-constraints.txt
```diff
@@ -12,3 +12,6 @@ WTforms<2.3.0
 
 # Details: https://github.com/apache/airflow/issues/8599
 flask-appbuilder<2.3.3
+
+# Avoid Version check if Version Check Table doesn't exist
+astronomer-airflow-version-check~=1.0.1
```
ashb, line 17: This would only install 1.0.*, but not a 1.1.

kaxil: Right yes.

1.10.10/alpine3.10/include/pip-constraints.txt
```diff
@@ -12,3 +12,6 @@ WTforms<2.3.0
 
 # Details: https://github.com/apache/airflow/issues/8599
 flask-appbuilder<2.3.3
+
+# Avoid Version check if Version Check Table doesn't exist
+astronomer-airflow-version-check>=1.0, <2
```
ashb, line 17: ITYM 
```suggestion
astronomer-airflow-version-check>=1.0.1, <2
```

.circleci/bin/test-airflow-image.py
```diff
@@ -85,7 +90,21 @@ def test_redis_version(webserver):
...
+def test_astronomer_airflow_check_version(webserver):
+    """ astronomer-airflow-version-check 1.0.0 has an issue in the Astronomer platform
+    """
+    try:
+        version_check_module = webserver.pip_package.get_packages()['astronomer-airflow-version-check']
+    except KeyError:
+        print("astronomer-airflow-version-check pip module is not installed")
+        return
+    version = version_check_module['version']
+    assert semantic_version(version) > semantic_version('1.0.0'), \
```
ashb, line 105:
```suggestion
    assert semantic_version(version) >= semantic_version('1.0.1'), \
```

Captures the intent better I feel.

PR #77, opened 2020-05-12
Constrain version of azure-storage

kaxil:
2020.05.12 release of the azure-storage is not installable any more
(it is deprecated). For now, we should switch to the latest working
version


kaxil:
I am merging this but will cancel the tests

PR #82, opened 2020-05-16
Push images with Build Number to DEV Docker Repo

kaxil:
I don't see a reason why we should publish images like `astronomerinc/ap-airflow:1.10.5-rhel7-onbuild-11056`

Let me know if there is a use-case for this, in which case I will add logic to push them to `astronomerio` repo


ashb:
I think having every merge to master going to our Dev docker repo might be useful? Then we could test builds/changes?

kaxil:
> I think having every merge to master going to our Dev docker repo might be useful? Then we could test builds/changes?

Yea that makes sense. Updated the PR

PR #86, opened 2020-05-26
Fix RHEL Image

kaxil:
Fixing this image till we decide to remove it (@schnie Happy for us to remove the image ? )


schnie:
@ryw any thoughts on rhel image here?

PR #90, opened 2020-06-02
Update AC 1.10.5 Alpine image to install Python 3.7.7

kaxil:
Added a variable for DEV Images Whitelisting for AC Versions that don't have Python Wheels

Created two tags in https://github.com/astronomer/airflow, that are same as 1.10.5-7
https://github.com/astronomer/airflow/releases/tag/1.10.5-8.dev
https://github.com/astronomer/airflow/releases/tag/1.10.5-8

### DIFF
--- a/.circleci/generate_circleci_config.py
+++ b/.circleci/generate_circleci_config.py
@@ -11,16 +11,27 @@ import re
 from jinja2 import Environment, FileSystemLoader
 
 IMAGE_MAP = collections.OrderedDict([
-    ("1.10.5-7", ["alpine3.10", "buster", "rhel7"]),
+    ("1.10.5-8.dev", ["alpine3.10", "buster", "rhel7"]),
     ("1.10.6-3", ["alpine3.10", "buster"]),
     ("1.10.7-11", ["alpine3.10", "buster"]),
     ("1.10.10-1", ["alpine3.10", "buster"]),
 ])
 
+# Airflow Versions for which we don't publish Python Wheels
+DEV_WHITELIST = ["1.10.5", "1.10.6"]
+
 
 def dev_releases(all_releases):
     """Find dev releases from a list of releases"""
-    return [release for release in all_releases if "dev" in release]
+    return [
+        release for release in all_releases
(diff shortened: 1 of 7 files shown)


1.10.5/alpine3.10/Dockerfile
```diff
@@ -105,7 +104,7 @@ RUN echo https://github.com/astronomer/ap-airflow/raw/${REPO_BRANCH}/alpine-pack
 	&& apk del .build-deps py3-numpy-dev \
 	&& ln -sf /usr/bin/python3 /usr/bin/python \
 	&& ln -sf /usr/bin/pip3 /usr/bin/pip \
-	&& sed -i -e '\#https://github.com/astronomer/astronomer/#d' /etc/apk/repositories
+	&& sed -i -e '\#https://github.com/astronomer/#d' /etc/apk/repositories
```
kaxil, line 107: I still don't know what this change was for in the original PR

.circleci/generate_circleci_config.py
```diff
@@ -11,16 +11,27 @@
...
 IMAGE_MAP = collections.OrderedDict([
-    ("1.10.5-7", ["alpine3.10", "buster", "rhel7"]),
+    ("1.10.5-8.dev", ["alpine3.10", "buster", "rhel7"]),
     ("1.10.6-3", ["alpine3.10", "buster"]),
     ("1.10.7-11", ["alpine3.10", "buster"]),
     ("1.10.10-1", ["alpine3.10", "buster"]),
 ])
 
+# Airflow Versions for which we don't publish Python Wheels
+DEV_WHITELIST = ["1.10.5", "1.10.6"]
```
kaxil, line 21: For the versions listed here:

- We won't create Nightly builds
- We won't look for the latest versions in pip.astronomer.io
- We would look for the GIT TAG matching the AC Version

1.10.5/alpine3.10/Dockerfile
ashb: The previous one didn't actually remove the line, so our registry was left configured, but it shouldn't be as we don't have a guaranteed stable signing key, and if it changes apk would break

PR #97, opened 2020-06-05
Add 1.10.10-3.dev

kaxil:
Only change:

- [[AIRFLOW-6959] Use NULL as dag.description default value (apache#7593)](https://github.com/astronomer/airflow/commit/548f4be32dbd4d222eb5666e83a9629ef360610e)


kaxil:
@ashb can you please take a look at this when you have time

PR #99, opened 2020-06-10
Test Onbuild Docker Images

kaxil:
We were previously not testing Onbuild images.

This PR:

- Use a single CircleCI step to test both base & onbuild images 
- Removes unused code


kaxil:
@ashb  I have already tested this PR: https://app.circleci.com/pipelines/github/astronomer/ap-airflow/533/workflows/f6f021ae-9ac2-470c-adcc-0299a96dac00

ashb:
Does it need to be two separate steps? Can't it be a single step in the pipeline, and a single`pytest` invocation that tests both images. (It's one build step, so one test step makes sense to me)

(This also makes the change smaller and easier to review)

kaxil:
> Does it need to be two separate steps? Can't it be a single step in the pipeline, and a single`pytest` invocation that tests both images. (It's one build step, so one test step makes sense to me)
> 
> (This also makes the change smaller and easier to review)

Sure, will do

kaxil:
@ashb Ready for review

.circleci/bin/test-airflow-image.py
```diff
@@ -195,72 +207,38 @@ def scheduler(request):
 
 @pytest.fixture(scope='session')
 def docker_client(request):
-    """ This is a text fixture for the docker client,
-    should it be needed in a test
-    """
+    """ This is a text fixture for the docker client, should it be needed in a test """
```
ashb, line 210:
```suggestion
    """ This is a test fixture for the docker client, should it be needed in a test """
```

.circleci/config.yml.j2
```diff
@@ -111,9 +110,8 @@ workflows:
             - build-{{ airflow_version }}-{{ distribution }}
       - test:
           name: test-{{ airflow_version }}-{{ distribution }}-onbuild
```
ashb, line 112:
```suggestion
          name: test-{{ airflow_version }}-{{ distribution }}-images
```

kaxil: Fixed

PR #111, opened 2020-06-23
Add script to automatically repackage apache_airflow_backport_providers

ashb:
We need to update their dependency from apache-airflow to
astronomer-certified to make them work with AC.

#### Checklist

### DIFF
--- a/repackage-providers.py
+++ b/repackage-providers.py
@@ -102,9 +102,6 @@ def update_metadata(unpacked_folder, ver: str):
 
     new_metadata.set_payload(metadata.get_payload())
 
-    #import difflib, sys
-    #sys.stdout.writelines(difflib.context_diff(metadata.as_string().splitlines(keepends=True), new_metadata.as_string().splitlines(keepends=True)))
-
     with open(os.path.join(dist_info_path, 'METADATA'), 'w') as fh:
         fh.write(new_metadata.as_string())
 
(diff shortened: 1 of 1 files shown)


ashb:
> Should this file live here or astronomer/airflow ?

The reason I did it here as it's not tied to any particular Airflow version -- so putting it in astronomer/airflow would mean it would have to live in v1-10-10, say, which feels odd.

repackage-providers.py
```diff
@@ -0,0 +1,156 @@
...
+            val = re.sub(r'^apache-airflow(\s|$)', r'astronomer-certified\1', val)
+        if key == 'Version':
+            val = ver
+
+        new_metadata[key] = val
+
+    new_metadata.set_payload(metadata.get_payload())
+
+    #import difflib, sys
+    #sys.stdout.writelines(difflib.context_diff(metadata.as_string().splitlines(keepends=True), new_metadata.as_string().splitlines(keepends=True)))
```
kaxil, line 106: This can go

repackage-providers.py
```diff
@@ -0,0 +1,156 @@
...
+    if metadata_ver != ver:
+        # Update the version in the .dist-info/ folder name, as this is what
+        # `wheel pack` uses to create the filename
+        folder = os.path.basename(dist_info_path)
+        folder = folder.replace(metadata_ver, ver)
+        os.rename(dist_info_path, os.path.join(unpacked_folder, folder))
+
+    return metadata['Name']
+
+#repack_wheel('https://dist.apache.org/repos/dist/dev/airflow/backport-providers/2020.6.24rc1/apache_airflow_backport_providers_amazon-2020.6.24rc1-py3-none-any.whl')
```
kaxil, line 120: And this too

PR #113, opened 2020-07-13
Add AC 1.10.11-1.dev

kaxil:
Create a nightly image of Airflow 1.10.11 image

1.10.11 Branch: https://github.com/astronomer/airflow/commits/v1-10-11

#### Checklist

- [x] If a new distribution or Airflow version is added, please add in .circleci/generate_circleci_config.py and run the script
- [x] If a new distribution or Airflow version is added, there Dockerfile in all base image directories
- [x] If a new distribution is added, it is supported by all Airflow versions
- [x] If a new Airflow version is added, it supports all distributions
- [x] If changing an image, add applicable test in .circleci/test-airflow-image.py


kaxil:
Thankfully our tests helped discover a bug :)

1.10.11/alpine3.10/Dockerfile
```diff
@@ -0,0 +1,123 @@
...
+	&& cp /usr/share/zoneinfo/UTC /etc/localtime \
+	&& pip3 install --no-cache-dir --upgrade snowflake-connector-python==1.9.1 \
+	&& pip3 install --no-cache-dir "${AIRFLOW_MODULE}" \
+	&& pip3 install --no-cache-dir "https://github.com/astronomer/astronomer-airflow-scripts/releases/download/v0.0.4/astronomer_airflow_scripts-0.0.4-py3-none-any.whl" \
+	&& pip3 install --no-cache-dir "astronomer-fab-security-manager~=1.2, >=1.2.2" \
+	&& apk del .build-deps py3-numpy-dev \
+	&& ln -sf /usr/bin/python3 /usr/bin/python \
+	&& ln -sf /usr/bin/pip3 /usr/bin/pip \
+	&& sed -i -e '\#https://github.com/astronomer/#d' /etc/apk/repositories \
+	&& sed -i -e 's/^run_as_user = .*/run_as_user = 100/g' /usr/lib/python3.7/site-packages/airflow/config_templates/default_airflow.cfg
```
kaxil, line 102: @schnie We change the default UID for alpine image here to `100` 

The tests are passing :)

ashb: I think I'd do this in a separate layer.

Also, we should probably set rbac=True default at the same time (Or have we done that in our wheels already?)

kaxil: `rbac=True` is done in our Helm chart: https://github.com/astronomer/airflow-chart/blob/master/templates/configmap.yaml#L34

kaxil: and moved the `sed` to separate layer

PR #115, opened 2020-07-21
Run Kubernetes Worker Pods as astro user

kaxil:
fixes https://github.com/astronomer/issues/issues/1572

### DIFF
--- a/1.10.5/rhel7/Dockerfile
+++ b/1.10.5/rhel7/Dockerfile
@@ -4,7 +4,7 @@ MAINTAINER Astronomer <humans@astronomer.io>
 LABEL maintainer="Astronomer <humans@astronomer.io>"
 
 LABEL io.astronomer.docker=true
-LABEL io.astronomer.docker.distro="alpine"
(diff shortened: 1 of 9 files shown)


1.10.5/rhel7/Dockerfile
```diff
@@ -4,7 +4,7 @@ MAINTAINER Astronomer <humans@astronomer.io>
 LABEL maintainer="Astronomer <humans@astronomer.io>"
 
 LABEL io.astronomer.docker=true
-LABEL io.astronomer.docker.distro="alpine"
+LABEL io.astronomer.docker.distro="rhel"
```
kaxil, line 7: I think this was a copy/paste error from whoever donated the RHEL dockerimage. 

cc @schnie

kaxil:
> LGTM, what happens when we add a new AC version? This will have to go along for the ride right?

Yeah, it will have to. We generally just copy-paste an old version folder and just replace versions so that would come along :)

PR #124, opened 2020-08-14
Restrict pinned Airflow version in the requirements.txt

kaxil:
Restrict Usage of the following patterns:

["apache-airflow==1.10.10", "apache-airflow>=1.10.5", "apache-airflow~=1.10.7", "apache-airflow == 1.10.10",]

### DIFF
--- a/.circleci/bin/test-airflow-image.py
+++ b/.circleci/bin/test-airflow-image.py
@@ -217,7 +217,8 @@ def test_apache_airflow_in_requirements(tmp_path):
     image_name = get_image_name(ImageType.ONBUILD.value)
 
     restricted_value_patterns = [
-        "apache-airflow==1.10.10", "apache-airflow>=1.10.5", "apache-airflow~=1.10.7"
+        "apache-airflow==1.10.10", "apache-airflow>=1.10.5", "apache-airflow~=1.10.7",
+        "apache-airflow == 1.10.10",
     ]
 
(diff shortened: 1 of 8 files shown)


.circleci/bin/test-airflow-image.py
```diff
@@ -216,12 +216,18 @@ def test_apache_airflow_in_requirements(tmp_path):
     test_project.mkdir()
     image_name = get_image_name(ImageType.ONBUILD.value)
 
+    restricted_value_patterns = [
+        "apache-airflow==1.10.10", "apache-airflow>=1.10.5", "apache-airflow~=1.10.7"
```
ashb, line 220: Try with spaces around the operator too?

kaxil: Done in https://github.com/astronomer/ap-airflow/pull/124/commits/a2ae1b6794968818bd989a2e438cc0850b0c4ca4

paolaperaza:
@kaxil Do you think we'll have to update this every time a minor/major version of Airflow comes out?

kaxil:
> @kaxil Do you think we'll have to update this every time a minor/major version of Airflow comes out?

No not at all, it uses the following regex:

https://github.com/astronomer/ap-airflow/blob/d74418ab1f0835660f7fb2655193b7b912e4356e/common/Dockerfile.onbuild-rhel7#L33-L35

The 1.10.10 example is just used for testing

PR #128, opened 2020-08-26
Add pinned constraints for 1.10.12 from 1.10.12.dev1 images

kaxil:
To ensure we test and deploy the same image, we should pin all the requirements.

Generated using the following:

```
docker run astronomerio/ap-airflow:1.10.12-alpine3.10-onbuild-16448 -- /entrypoint pip3 freeze > 1.10.12/alpine3.10/include/pip-constraints.txt

docker run astronomerio/ap-airflow:1.10.12-buster-onbuild-16446 -- /entrypoint pip3 freeze > 1.10.12/buster/include/pip-constraints.txt
```

Part of https://github.com/astronomer/issues/issues/1271


kaxil:
WDYT of this @ashb ? Just wanted to get an opinion

ashb:
Hmmm, we probably want this for build time, but I'm not sure if we want the full pin to affect what is installed via requirements.txt.

Can't quite work though the implications of this change

PR #129, opened 2020-08-27
Use role_based_auth as default auth_backend for AC 1.10.12

kaxil:
Commit: https://github.com/astronomer/airflow/commit/49d4840b0058e673e4d698f16bc6f4fac6eea9d4

### DIFF
--- a/1.10.12/alpine3.10/Dockerfile
+++ b/1.10.12/alpine3.10/Dockerfile
@@ -101,7 +101,9 @@ RUN echo https://github.com/astronomer/ap-airflow/raw/${REPO_BRANCH}/alpine-pack
 	&& sed -i -e '\#https://github.com/astronomer/#d' /etc/apk/repositories
 
 # Run pods spun up by Kubernetes Executor as astro user
-RUN sed -i -e 's/^run_as_user =.*/run_as_user = 100/g' /usr/lib/python3.7/site-packages/airflow/config_templates/default_airflow.cfg
+# And use role_based_auth as auth_backend
+RUN sed -i -e 's/^run_as_user =.*/run_as_user = 100/g; s/^auth_backend =.*/auth_backend = airflow.api.auth.backend.role_based_auth/g' \
+    /usr/lib/python3.7/site-packages/airflow/config_templates/default_airflow.cfg
 
 # Create logs directory so we can own it when we mount volumes
 RUN mkdir -p ${AIRFLOW_HOME}/logs
--- a/1.10.12/buster/Dockerfile
+++ b/1.10.12/buster/Dockerfile
@@ -139,7 +139,9 @@ COPY include/pip.conf /etc/pip.conf
 COPY include/pip-constraints.txt /usr/local/share/astronomer-pip-constraints.txt
 
 # Run pods spun up by Kubernetes Executor as astro user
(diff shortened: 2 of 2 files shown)


1.10.12/alpine3.10/Dockerfile
```diff
@@ -101,7 +101,9 @@ RUN echo https://github.com/astronomer/ap-airflow/raw/${REPO_BRANCH}/alpine-pack
 	&& sed -i -e '\#https://github.com/astronomer/#d' /etc/apk/repositories
 
 # Run pods spun up by Kubernetes Executor as astro user
-RUN sed -i -e 's/^run_as_user =.*/run_as_user = 100/g' /usr/lib/python3.7/site-packages/airflow/config_templates/default_airflow.cfg
+# And use role_based_auth as auth_backend
+RUN sed -i -e 's/^run_as_user =.*/run_as_user = 100/g; s/^auth_backend =.*/auth_backend = airflow.api.auth.backend.role_based_auth/g' \
+    /usr/lib/python3.7/site-packages/airflow/config_templates/default_airflow.cfg
```
kaxil, line 106: We could add this to our Helm Chart too but there we will need to verify the AC Version. Even if we backport this to all previous AC Versions we will need to check if all users are using the latest AC images

schnie: Would we not be able to add these changes into https://github.com/astronomer/airflow-chart/blob/master/templates/configmap.yaml#L19?

kaxil: Yes that is definitely one option. I didn't do it there is not all of our Airflow deployment will have this auth backend (commit: https://github.com/astronomer/airflow/commit/49d4840b0058e673e4d698f16bc6f4fac6eea9d4), currently, we have just added it to **1.10.12-1.dev**. Even if we added it to AC 1.10.5 , 1.10.7 and 1.10.10 , not all the deployments would be using the new updated image with the auth_backend so it will causes error.

ashb: Could we not just change the default in default_airflow.cfg -- i.e in https://github.com/astronomer/airflow? (or similar)? Is there a reason we haven't done that already?

kaxil: I didn't do it over there as the AC users (not using the platform) won't have our Astro FAB Security Manager. And without that security manager the endpoints remain "open"

ashb: People can use our docker images outside of the platform too thought

kaxil: The FAB Security Manager (https://github.com/astronomer/ap-airflow/blob/master/1.10.12/alpine3.10/Dockerfile#L97) is installed in the Dockerfile at the very least so I placed it here :)

ashb: It's installed, but only used here: https://github.com/astronomer/airflow-chart/blob/a7598890f8f7ea6f1df80fdf705fcaa3f345b54a/templates/configmap.yaml#L116-L121

I'm not sure we can do the same easily.

That said, if we did make this the default, it would have the same default of "deny_all", as without our auth/JWT layer in front, it wouldn't ever find a valid user (unless made via a logged in browser session etc.)

Not sure -- still operating on about 20% brain capacity so do what you think is best.

kaxil:
Closing this in favor of https://github.com/astronomer/airflow/pull/878

PR #131, opened 2020-09-09
Add AC 1.10.7-15.dev and 1.10.10-5.dev

kaxil:
https://github.com/astronomer/airflow/commits/v1-10-10
https://github.com/astronomer/airflow/commits/v1-10-7


ernest-kr (approved):
Approving based on Kaxil

PR #133, opened 2020-09-17
Add astro user to group 101

ianstanton:
- When using an IAM role in place of connection credentials + our debian image, we see the following error:
`[Errno 13] Permission denied: '/var/run/secrets/eks.amazonaws.com/serviceaccount/token'`
- Becuase we run with `uid 100` and `gid 101` by default (in airflow-chart), `/var/run/secrets/eks.amazonaws.com/serviceaccount/` has the following permissions set and is accessible by `gid 101`:
  - `drwxrwsrwt 3 root 101 100 Sep 15 18:36 serviceaccount`
- We need to add our `astro` (which by default has `uid 50000` and `gid 50000`) to `gid 101`
- Related to https://github.com/astronomer/issues/issues/1797

- The group `101` doesn't actually exist and shows up as `nogroup`. I added a step to create this group and named it `astrogroup`. Please feel free to change this name.
#### Checklist

### DIFF
--- a/1.10.10/buster/Dockerfile
+++ b/1.10.10/buster/Dockerfile
@@ -77,7 +77,10 @@ RUN apt-get update \
 
 RUN pip install --upgrade pip==19.3.1
 
-RUN useradd --uid $ASTRONOMER_UID --create-home ${ASTRONOMER_USER}
+RUN useradd --uid $ASTRONOMER_UID --create-home ${ASTRONOMER_USER} \
+    && groupadd astrogroup --gid 101 \
+    && usermod --append --groups astrogroup ${ASTRONOMER_USER}
+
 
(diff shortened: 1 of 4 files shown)


ashb (approved):

kaxil (commented):
We would want to make this change in all the buster images, right @ianstanton ?

i.e. 1.10.5, 1.10.7 & 1.10.12-dev buster images too

ianstanton:
@kaxil Yes, good catch. I'll add to those as well.

kaxil (approved):
🚀 Thanks @ianstanton

vishwas-astro:
@ianstanton the backport here is basically to support the AC images that this bug fix gets into in the platform. Correct?

PR #140, opened 2020-10-12
Add 2.0.0-1.dev image

kaxil:
https://github.com/astronomer/airflow/tree/v2-0-0

### DIFF
--- a/2.0.0/buster/include/clean-airflow-logs
+++ b/2.0.0/buster/include/clean-airflow-logs
@@ -1,0 +1,20 @@
+#!/usr/bin/env bash
+
+set -e
+
+DIRECTORY=${AIRFLOW_HOME:-/usr/local/airflow}
+RETENTION=${ASTRONOMER__AIRFLOW__WORKER_LOG_RETENTION_DAYS:-15}
+
+trap "exit" INT TERM
+
+EVERY=$((15*60))
+
+echo "Cleaning logs every $EVERY seconds"
+
+while true; do
+  seconds=$(( $(date -u +%s) % EVERY))
+  [[ $seconds -lt 1 ]] || sleep $((EVERY - seconds))
+
+  echo "Trimming airflow logs to ${RETENTION} days."
+  find "${DIRECTORY}"/logs -mtime +"${RETENTION}" -name '*.log' -delete
+done
--- a/common/Dockerfile.onbuild-buster
+++ b/common/Dockerfile.onbuild-buster
@@ -16,6 +16,7 @@
 ARG baseimage=astronomerinc/ap-airflow:1.10.7-buster
 
 FROM ${baseimage}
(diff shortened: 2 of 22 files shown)


kaxil:
This PR is blocked by Helm Chart support for 2.0, merging https://github.com/astronomer/airflow-chart/pull/122 and https://github.com/astronomer/airflow-chart/pull/125 and releasing a new version with it should then make the tests pass for this PR.

cc @schnie @vishwas-astro

2.0.0/buster/include/clean-airflow-logs
```diff
@@ -0,0 +1,20 @@
...
+
+EVERY=$((15*60))
+
+echo "Cleaning logs every $EVERY seconds"
+
+while true; do
+  seconds=$(( $(date -u +%s) % EVERY))
+  [[ $seconds -lt 1 ]] || sleep $((EVERY - seconds))
+
+  echo "Trimming airflow logs to ${RETENTION} days."
```
ashb, line 18: We don't need this script for 2.0 onwards I think as it's upstreamed.

Let's check that later

ashb (approved):

common/Dockerfile.onbuild-buster
```diff
@@ -16,6 +16,7 @@
 ARG baseimage=astronomerinc/ap-airflow:1.10.7-buster
 
 FROM ${baseimage}
+SHELL ["/bin/bash", "-o", "pipefail", "-e", "-u", "-x", "-c"]
```
kaxil, line 19: I am not 100% sure on why this is needed on why I had to explicitly set it here since the base image too has this set but without it, I got the following error and the packages (like git) were not installed:

https://github.com/astronomer/issues/issues/1964

```
❯ astro dev start
Env file ".env" found. Loading...
Sending build context to Docker daemon    191kB
Step 1/4 : FROM astronomerio/ap-airflow:2.0.0-1.dev3-buster-onbuild
# Executing 7 build triggers
 ---> Running in 5d913c3de7a4
Removing intermediate container 5d913c3de7a4
 ---> Running in d65158eab62c
/bin/sh: 1: [[: not found
Removing intermediate container d65158eab62c
 ---> Running in 17c8fc558cd3
```

kaxil: cc @ashb If you have any ideas on why this would have occurred. In any case, this should be harmless

ashb: I suspect it might not propogate from three base image as it's build time only

ashb (approved):

dimberman (approved):

kaxil (commented):
This PR is currently blocked by https://github.com/astronomer/airflow-chart/pull/133 and https://github.com/astronomer/issues/issues/2007

kaxil:
Merging since 2.0.0 test passed but 1.10.12 test has failed so we need to check more on that

PR #142, opened 2020-10-24
Improve ci

danielhoherd:

These are basically no-op changes that make the .circleci/config.yml file a bit more readable and easier to parse. This also bumps the machine image to ubuntu 20.04 and the python version to 3.8.5, and adds some quotes around some variables that could cause problems.

These are CI changes only and should be a no-op for existing features and functions, but should all still work on the newer versions of ubuntu and python.

I'd like to have these changes in before working on pushing to quay.io, just because adding these and the quay changes was creating a huge change set of cosmetic and version changes in here and the feature changes of pushing to multiple repos.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -740,7 +740,7 @@ jobs:
       - airflow-image-test:
           repository: "<< parameters.docker_repo >>"
           tag: "<< parameters.tag >>"
-  scan:
+  scan-clair:
     executor: clair-scanner/default
     description: Test Airflow images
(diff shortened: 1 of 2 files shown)


kaxil:
Looks like merging https://github.com/astronomer/ap-airflow/pull/138 caused some conflicts

danielhoherd:
@kaxil all fixed.

kaxil (approved):

kaxil:
Thanks @danielhoherd

PR #145, opened 2020-10-29
Rework CI and push to quay.io

danielhoherd:
## Description

- Push to quay.io and docker hub
- Rework a lot of this flow, because it was difficult to work the dual push in without breaking some of the existing code assumptions

## Notes to the reviewers

This is a complicated, big change set, so I apologize for that. Grokking the flow through the templating and CI flows took a lot of time, and while I think I managed to handle all of the desired behaviors, it's not unreasonable that I will have done something wrong that needs to be fixed, or slips through and begins doing incorrect builds and pushes in the future.

If anybody wants to zoom chat about these changes we can do that. If we want to rework this to a smaller change set, or to find other ways to simplify this CI pipeline, we can discuss that too.

I recommend looking at the j2 file first, and also there are some past CI pipelines with logs to inspect to verify that things are performing the desired steps.


kaxil (approved):

kaxil:
Thanks @danielhoherd , amazing effort, LGTM

PR #146, opened 2020-10-29
Stopping running sync_perm multiple times for Airflow >= 1.10.7

kaxil:
We already run `sync_perm` in the entrypoint, so sync'ing permissions when the webserver is started is not needed. Otherwise each gunicorn workers run this commands unncessary and cause a delayed start

closes https://github.com/astronomer/issues/issues/1962

### DIFF
--- a/.circleci/bin/test-airflow-image.py
+++ b/.circleci/bin/test-airflow-image.py
@@ -251,6 +251,12 @@ def test_airflow_configs(scheduler, docker_client):
                 "cat /usr/lib/python3.7/site-packages/airflow/config_templates/default_airflow.cfg | "
                 "grep '^run_as_user' | awk '{print $3}'").strip() == ""
 
+    if semantic_version(airflow_version) >= semantic_version('1.10.10'):
+        assert scheduler.check_output(
+            "cat /usr/local/lib/python3.7/site-packages/airflow/config_templates/default_airflow.cfg | "
+            "grep '^update_fab_perms' | awk '{print $3}'"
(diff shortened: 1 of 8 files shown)


.circleci/bin/test-airflow-image.py
```diff
@@ -251,6 +251,12 @@ def test_airflow_configs(scheduler, docker_client):
                 "cat /usr/lib/python3.7/site-packages/airflow/config_templates/default_airflow.cfg | "
                 "grep '^run_as_user' | awk '{print $3}'").strip() == ""
 
+    if semantic_version(airflow_version) >= semantic_version('1.10.10'):
```
kaxil, line 254: This flag is only available from 1.10.7

https://airflow.readthedocs.io/en/latest/configurations-ref.html#update-fab-perms

.circleci/bin/test-airflow-image.py
```diff
@@ -251,6 +251,12 @@ def test_airflow_configs(scheduler, docker_client):
                 "cat /usr/lib/python3.7/site-packages/airflow/config_templates/default_airflow.cfg | "
                 "grep '^run_as_user' | awk '{print $3}'").strip() == ""
 
+    if semantic_version(airflow_version) >= semantic_version('1.10.7'):
```
kaxil, line 254: This flag is only available from 1.10.7

https://airflow.readthedocs.io/en/latest/configurations-ref.html#update-fab-perms

ashb (approved):

ashb (approved):

PR #153, opened 2020-11-16
Add four labels to 'docker build'

danielhoherd:
## Description

Add four labels to every `docker build` step. This will make it easier to track the source and history of built images.


danielhoherd:
@kaxil looks like we can merge this now.

kaxil (approved):

PR #157, opened 2020-11-18
Don't check for commits against master -- it breaks CI

ashb:
We can't have this check -- as it breaks when pre-commit is used in CI
to run the same checks.

#### Checklist


ashb:
We've got protected branches set up in Github anyway, so the push will be denied.

kaxil (approved):

PR #160, opened 2020-11-24
Use Airflow Constraints file for AC 1.10.12 and 1.10.13

kaxil:
Use Apache Airflow constraints file to maintain dependencies during build time for AC 1.10.12 and AC 1.10.13

### DIFF
--- a/1.10.10/alpine3.10/Dockerfile
+++ b/1.10.10/alpine3.10/Dockerfile
@@ -21,12 +21,13 @@ ARG VERSION="1.10.10-6.*"
 ARG SUBMODULES="all, statsd, elasticsearch"
 ARG AIRFLOW_MODULE="astronomer_certified[${SUBMODULES}]==$VERSION"
 ARG REPO_BRANCH=master
+ENV AIRFLOW_VERSION="1.10.10"
 
 LABEL io.astronomer.docker=true
 LABEL io.astronomer.docker.distro="alpine"
 LABEL io.astronomer.docker.module="airflow"
 LABEL io.astronomer.docker.component="airflow"
-LABEL io.astronomer.docker.airflow.version="1.10.10"
+LABEL io.astronomer.docker.airflow.version="${AIRFLOW_VERSION}"
 LABEL io.astronomer.docker.ac.version="${VERSION}"
 
 ENV AIRFLOW_HOME="/usr/local/airflow"
--- a/1.10.10/alpine3.10/Dockerfile
+++ b/1.10.10/alpine3.10/Dockerfile
@@ -21,7 +21,7 @@ ARG VERSION="1.10.10-6.*"
 ARG SUBMODULES="all, statsd, elasticsearch"
 ARG AIRFLOW_MODULE="astronomer_certified[${SUBMODULES}]==$VERSION"
 ARG REPO_BRANCH=master
-ENV AIRFLOW_VERSION="1.10.10"
+ARG AIRFLOW_VERSION="1.10.10"
 
 LABEL io.astronomer.docker=true
 LABEL io.astronomer.docker.distro="alpine"
--- a/1.10.12/alpine3.10/Dockerfile
+++ b/1.10.12/alpine3.10/Dockerfile
@@ -21,7 +21,7 @@ ARG VERSION="1.10.12-2.*"
 ARG SUBMODULES="all, statsd, elasticsearch"
 ARG AIRFLOW_MODULE="astronomer_certified[${SUBMODULES}]==$VERSION"
 ARG REPO_BRANCH=master
-ENV AIRFLOW_VERSION="1.10.12"
+ARG AIRFLOW_VERSION="1.10.12"
 
 LABEL io.astronomer.docker=true
 LABEL io.astronomer.docker.distro="alpine"
--- a/1.10.12/alpine3.10/Dockerfile
+++ b/1.10.12/alpine3.10/Dockerfile
@@ -21,12 +21,13 @@ ARG VERSION="1.10.12-2.*"
 ARG SUBMODULES="all, statsd, elasticsearch"
 ARG AIRFLOW_MODULE="astronomer_certified[${SUBMODULES}]==$VERSION"
 ARG REPO_BRANCH=master
+ENV AIRFLOW_VERSION="1.10.12"
 
 LABEL io.astronomer.docker=true
 LABEL io.astronomer.docker.distro="alpine"
 LABEL io.astronomer.docker.module="airflow"
 LABEL io.astronomer.docker.component="airflow"
-LABEL io.astronomer.docker.airflow.version="1.10.12"
+LABEL io.astronomer.docker.airflow.version="${AIRFLOW_VERSION}"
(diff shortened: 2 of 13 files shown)


1.10.10/alpine3.10/Dockerfile
```diff
@@ -21,12 +21,13 @@ ARG VERSION="1.10.10-6.*"
 ARG SUBMODULES="all, statsd, elasticsearch"
 ARG AIRFLOW_MODULE="astronomer_certified[${SUBMODULES}]==$VERSION"
 ARG REPO_BRANCH=master
+ENV AIRFLOW_VERSION="1.10.10"
```
ashb, line 24: Why make this an env?

1.10.12/alpine3.10/Dockerfile
```diff
@@ -21,12 +21,13 @@ ARG VERSION="1.10.12-2.*"
 ARG SUBMODULES="all, statsd, elasticsearch"
 ARG AIRFLOW_MODULE="astronomer_certified[${SUBMODULES}]==$VERSION"
 ARG REPO_BRANCH=master
+ENV AIRFLOW_VERSION="1.10.12"
```
ashb, line 24:
```suggestion
ARG AIRFLOW_VERSION="1.10.12"
```

Env is visible in the resulting docker image, but it doesn't need to be.

1.10.12/alpine3.10/Dockerfile
```diff
@@ -92,7 +93,7 @@ RUN echo https://github.com/astronomer/ap-airflow/raw/${REPO_BRANCH}/alpine-pack
 	&& update-ca-certificates \
 	&& cp /usr/share/zoneinfo/UTC /etc/localtime \
 	&& pip3 install --no-cache-dir --upgrade snowflake-connector-python==1.9.1 \
-	&& pip3 install --no-cache-dir "${AIRFLOW_MODULE}" \
+	&& pip3 install "${AIRFLOW_MODULE}" --constraint "https://raw.githubusercontent.com/apache/airflow/constraints-${AIRFLOW_VERSION}/constraints-3.7.txt" \
```
ashb, line 96: Does this overwrite-replace the constraint file we specify in /usr/local/share/pip.conf?

kaxil: No, it does not overwrite or replace. The one we specify is also used and is baked in the Dockerfile so users can not install dependencies that don't satisfy our constraints file.

While the OSS one is just used during build time.

1.10.12/alpine3.10/Dockerfile
kaxil: The only reason I did that was so no one can overwrite the Airflow Version during docker-build time.

I can change that to using ARG but just was overly-cautious :D

kaxil: updated in https://github.com/astronomer/ap-airflow/pull/160/commits/28339144b747aadea11ed8b215697b8acc4b805d

1.10.10/alpine3.10/Dockerfile
kaxil: Updated in updated in https://github.com/astronomer/ap-airflow/pull/160/commits/28339144b747aadea11ed8b215697b8acc4b805d

1.10.12/alpine3.10/Dockerfile
ashb: I mean does _this_ command just use the one from the command line, or do we combine them?

(Not sure it matters which behaviuor, just curious)

kaxil: It uses both.

As a test I added `azure-storage<0.37.0` to `azure-storage>0.37.0` in `1.10.13/buster/include/pip-constraints.txt`

```
Collecting azure-storage-blob==2.1.0
  Downloading https://files.pythonhosted.org/packages/3e/84/610f379b46d7d3c2d48eadeed6a12b6d46a43100fea70534f5992d0ac996/azure_storage_blob-2.1.0-py2.py3-none-any.whl (88kB)
ERROR: Could not find a version that satisfies the requirement azure-storage>0.37.0 (from -c /usr/local/share/astronomer-pip-constraints.txt (line 6)) (from versions: 0.20.0rc1, 0.20.0rc2, 0.20.0, 0.20.1, 0.20.2, 0.20.3, 0.30.0, 0.31.0, 0.32.0, 0.33.0, 0.34.0, 0.34.2, 0.34.3, 0.35.0, 0.35.1, 0.36.0, 0.37.0)
ERROR: No matching distribution found for azure-storage>0.37.0 (from -c /usr/local/share/astronomer-pip-constraints.txt (line 6))
WARNING: You are using pip version 19.3.1; however, version 20.2.4 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
The command '/bin/bash -o pipefail -e -u -x -c pip install "${AIRFLOW_MODULE}" --constraint "https://raw.githubusercontent.com/apache/airflow/constraints-${AIRFLOW_VERSION}/constraints-3.7.txt"   && pip install "https://github.com/astronomer/astronomer-airflow-scripts/releases/download/v0.0.5/astronomer_airflow_scripts-0.0.5-py3-none-any.whl" 	&& pip install "astronomer-fab-security-manager~=1.2, >=1.2.2"' returned a non-zero code: 1
```

kaxil: Looks like `pip` evaluates from left to right when multiples constraints are used. And if a dependency exist in both left one is used. In our case `pip.conf` is the default so is `left` and has precedence over the constraint file passed by Airflow which is good so that we can override some constraints if need be


```
❯ echo "attrs < 20.3.0" >> const.txt
❯ pip install attrs --constraint https://raw.githubusercontent.com/apache/airflow/constraints-1.10.13/constraints-3.7.txt -c const.txt
Requirement already satisfied: attrs==20.3.0 in /Users/kaxilnaik/opt/anaconda3/lib/python3.7/site-packages (from -c https://raw.githubusercontent.com/apache/airflow/constraints-1.10.13/constraints-3.7.txt (line 50)) (20.3.0)
❯ pip install attrs -c const.txt --constraint https://raw.githubusercontent.com/apache/airflow/constraints-1.10.13/constraints-3.7.txt
Collecting attrs<20.3.0 (from -c const.txt (line 2))
  Using cached https://files.pythonhosted.org/packages/14/df/479736ae1ef59842f512548bacefad1abed705e400212acba43f9b0fa556/attrs-20.2.0-py2.py3-none-any.whl
ERROR: dbt-core 0.16.0 has requirement requests<2.23.0,>=2.18.0, but you'll have requests 2.24.0 which is incompatible.
ERROR: apache-airflow 1.10.12 has requirement attrs~=19.3, but you'll have attrs 20.2.0 which is incompatible.
Installing collected packages: attrs
  Found existing installation: attrs 20.3.0
    Uninstalling attrs-20.3.0:
      Successfully uninstalled attrs-20.3.0
Successfully installed attrs-20.2.0
```

ashb: Oh ace!

ashb (approved):

PR #198, opened 2021-01-18
Approvals Needed by the Team to generate new Docker Builds 

ernest-kr:
Approvals Needed by the Team to generate new Docker Builds 
Now Notifications would be  sent to a specific Slack Channel for Build Approvals
Instead of generating docker images for all the releases everytime now we can restrict to a particular releases by manually approving

Please review the Jinja template which has majority of the changes.


kaxil (approved):
There are some conflicts you will need to resolve

PR #201, opened 2021-01-22
Fix pre-commit git diff exit code

danielhoherd:
Make CI exit non-zero when `pre-commit run --all-files` fails. The `|| git diff` was causing it to return 0.

### DIFF
--- a/.circleci/config.yml.j2
+++ b/.circleci/config.yml.j2
@@ -152,7 +152,7 @@ jobs:
           command: |
             pyenv global 3.8.5
             pip install -r .circleci/test-requirements.txt
-            pre-commit run --all-files || git --no-pager diff
+            pre-commit run --all-files || { git --no-pager diff && false ; }
   build:
     executor: docker-executor
     description: Build Airflow images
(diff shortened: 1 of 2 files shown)


.circleci/config.yml.j2
```diff
@@ -152,7 +152,7 @@ jobs:
           command: |
             pyenv global 3.8.5
             pip install -r .circleci/test-requirements.txt
-            pre-commit run --all-files || git --no-pager diff
+            pre-commit run --all-files || { git --no-pager diff && false ; }
```
kaxil, line 155: Wouldn't it be better?

```suggestion
            pre-commit run --all-files || { git --no-pager diff && exit 1 ; }
```

danielhoherd: In CI, if any command exits non-0, the test will fail. `false` itself exits 1, but `exit 1` will exit the running shell with 1, which is a different behavior because the shell is the parent process. `false` in this case is functionally identical to `pre-commit run --all-files` returning non-zero, whereas `exit 1` would be like `pre-commit run --all-files || exit 1`

kaxil: aah good point

kaxil (approved):

PR #206, opened 2021-01-29
Remove clair. Bump ubuntu and pre-commit.

danielhoherd:

We have removed clair from all our other repos already. Quay uses clair for its security scans, plus we have Snyk and Trivy. I wasn't sure if we wanted to remove it here since this overlaps with the community stuff quite a bit, but here is the PR just in case.

https://github.com/astronomer/issues/issues/2543

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -90,12 +90,6 @@ workflows:
           requires:
             - Need-Approval-1.10.5
             - static-checks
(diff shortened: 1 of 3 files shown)


kaxil:
Can you rebase on master, please 😄  -- you can remove the commented blocks that temporarily disabled clair

kaxil (approved):

PR #212, opened 2021-02-09
Constrain versions for Providers during docker build for 2.0.0

kaxil:
Since we are planning to cut 2.0.0-3 we should release the same set of Providers except for few exceptions:

`apache-airflow-providers-elasticsearch==1.0.1` -- contains our logging fix
`apache-airflow-providers-cncf-kubernetes==1.0.1 `-- contains a bug fix for K8s Operator
`apache-airflow-providers-microsoft-azure==1.1.0` -- Will allow installing snowflake provider (includes https://github.com/apache/airflow/pull/12188)

Since these constraint sare only used in docker build time and not included in Dockerfile users are free to upgrade to a major version for providers like google. This is the main different between this constraints file vs `pip-constraints.txt`.

Without this change, new versions of providers would be installed (even the major ones)

### DIFF
--- a/2.0.0/CHANGELOG.md
+++ b/2.0.0/CHANGELOG.md
@@ -9,6 +9,15 @@ Astronomer Certified 2.0.0-3.dev
 - BugFix: Dag-level Callback Requests were not run ([commit](https://github.com/astronomer/airflow/commit/4412aba55))
 - Increase the default ``min_file_process_interval`` to decrease CPU Usage ([commit](https://github.com/astronomer/airflow/commit/6de02157f))
 - Stop creating duplicate Dag File Processors ([commit](https://github.com/astronomer/airflow/commit/4ced807ab))
+- Fix drop_user_and_chart migration rule for MySQL ([commit](https://github.com/astronomer/airflow/commit/ebbbc62b542c32b791dd265338d253ed1c1c19f9))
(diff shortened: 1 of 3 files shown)


ernest-kr:
@kaxil There is a problem if users try to upgrade the azure provider from 1.0.0 to 1.1.0.

https://github.com/apache/airflow/pull/12188

This workaround will fix . Shall we include this ?

USER root
RUN pip uninstall azure-storage --yes
RUN pip install azure-storage-blob
USER astro

kaxil:
> @kaxil There is a problem if users try to upgrade the azure provider from 1.0.0 to 1.1.0.
> 
> [apache/airflow#12188](https://github.com/apache/airflow/pull/12188)
> 
> This workaround will fix . Shall we include this ?
> 
> USER root
> RUN pip uninstall azure-storage --yes
> RUN pip install azure-storage-blob
> USER astro

@ernest-kr the PR already fixes that

kaxil:
> @kaxil There is a problem if users try to upgrade the azure provider from 1.0.0 to 1.1.0.
> 
> [apache/airflow#12188](https://github.com/apache/airflow/pull/12188)
> 
> This workaround will fix . Shall we include this ?
> 
> USER root
> RUN pip uninstall azure-storage --yes
> RUN pip install azure-storage-blob
> USER astro

@ernest-kr the PR already fixes that, we don't use the old azure provider, check my PR Description

kaxil:
ping @ashb @ernest-kr

ashb (approved):
Hmm I'm not sure. L

PR #217, opened 2021-03-02
Lets users rebuild using different python versions in our image

cmarteepants:
This PR provides a straightforward mechanism for users to rebuild the Astro Certified Airflow image with different python versions.
When building a custom image, users can set build arg `PYTHON_MAJOR_MINOR_VERSION` with a different python version, such as 3.8 or 3.6.

For example: `docker build --build arg PYTHON_MAJOR_MINOR_VERSION=3.6 -t astronomer-certified:python3.6-2.0.0-buster`

Based on the arg, we will pull the appropriate constraint file (where applicable) and copy all python modules from the appropriate place.

### DIFF
--- a/1.10.12/buster/Dockerfile
+++ b/1.10.12/buster/Dockerfile
@@ -119,7 +119,7 @@ COPY include/pip.conf /etc/pip.conf
 COPY include/pip-constraints.txt /usr/local/share/astronomer-pip-constraints.txt
 
 # Pip install airflow and astro security manager
-RUN pip install "${AIRFLOW_MODULE}" --constraint "https://raw.githubusercontent.com/apache/airflow/constraints-${AIRFLOW_VERSION}/constraints-${PYTHON_MAJOR_MINOR_VERSION}.txt" \
+RUN pip install "${AIRFLOW_MODULE}" --constraint "https://raw.githubusercontent.com/apache/airflow/constraints-${AIRFLOW_VERSION}/constraints-${PYTHON_MAJOR_MINOR_VERSION}.txt" \
   && pip install "https://github.com/astronomer/astronomer-airflow-scripts/releases/download/v0.0.5/astronomer_airflow_scripts-0.0.5-py3-none-any.whl" \
 	&& pip install "astronomer-fab-security-manager~=1.2, >=1.2.2"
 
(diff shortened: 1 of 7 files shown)


kaxil (approved):
Looks good, thanks @cmarteepants  -- Separate issue: we should probably document how to build the docker image with different args 

cc @paolaperaza

1.10.12/buster/Dockerfile
```diff
@@ -116,7 +119,7 @@ COPY include/pip.conf /etc/pip.conf
 COPY include/pip-constraints.txt /usr/local/share/astronomer-pip-constraints.txt
 
 # Pip install airflow and astro security manager
-RUN pip install "${AIRFLOW_MODULE}" --constraint "https://raw.githubusercontent.com/apache/airflow/constraints-${AIRFLOW_VERSION}/constraints-3.7.txt" \
+RUN pip install "${AIRFLOW_MODULE}" --constraint "https://raw.githubusercontent.com/apache/airflow/constraints-${AIRFLOW_VERSION}/constraints-${PYTHON_MAJOR_MINOR_VERSION}.txt" \
```
kaxil, line 122:
```suggestion
RUN pip install "${AIRFLOW_MODULE}" --constraint "https://raw.githubusercontent.com/apache/airflow/constraints-${AIRFLOW_VERSION}/constraints-${PYTHON_MAJOR_MINOR_VERSION}.txt" \
```

cmarteepants: Oh man, bad copy/paste! On it

ernest-kr:
@kaxil Dont we have to change the config.yaml since we need to pass the python version to build the image

kaxil:
> @kaxil Dont we have to change the config.yaml since we need to pass the python version to build the image

Since the default is already set in the dockerfile we don't need to. We only need to if we want to override the default arguments.

For example, this now allows us to publish docker images with multiple python version, so if we plan to publish images with 3.6 and 3.8 too, we just need to add some steps to config.yaml and pass `--build arg PYTHON_MAJOR_MINOR_VERSION="3.6"` or `--build arg PYTHON_MAJOR_MINOR_VERSION="3.8"`

PR #221, opened 2021-03-10
Fix 1.10.15-1.dev

kaxil:
- 1.10.15-1 is not released yet and is only a dev version
- nightly build should be created for this version


danielhoherd (approved):
Apologies for missing this detail on my last review.

PR #222, opened 2021-03-10
Sync fab perms on web startup using airfl config

cmarteepants:

Because of the custom fab security manager, permissions need to be synced in Airflow on webserver startup. This has up until now been done in the `entrypoint.sh` script.

Now that we are actively promoting our Astronomer Core image and promising it to be the same (but better!) as the OSS one, we'll hopefully see more users attempting to quickly run our containers. It's probable that some of these users will be evaluating the feasibility of migrating from whatever they are using to ours.

Something a lot of people do when first evaluating a container is to pull it and run it in interactive mode. This action overrides
the docker entrypoint, which means the permissions don't get synced. When they run the webserver and login, they are
stuck in an infinite loop. It's won't be apparent as to why, especially since they are likely not to know about the Astro custom fab security manager. If they then try the same action in the OSS image, it will work and they are likely not to return to our image. 

There is an Airflow configuration called `update_fab_perms` that can sync permissions automatically when the webserver starts. Easiest way to provide flexibility and meet users where they are without losing features when using gen1 platform is to leverage this instead.

On a not so random note: The `entrypoint.sh` script can also easily get overridden when people quickly try to scaffold an environment for an Airflow ldap integration tutorial 🤣 This may or may not have happened to me this morning. But it's mostly about the first thing!

#### Checklist

### DIFF
--- a/1.10.10/buster/Dockerfile
+++ b/1.10.10/buster/Dockerfile
@@ -152,8 +152,6 @@ COPY include/pip-constraints.txt /usr/local/share/astronomer-pip-constraints.txt
 RUN sed -i \
     # Run pods spun up by Kubernetes Executor as astro user
     -e 's/^run_as_user =.*/run_as_user = 50000/g' \
-    # We sync permissions in the entrypoint so do not need to run in the Webserver again
-    -e 's/^update_fab_perms =.*/update_fab_perms = False/g' \
     /usr/local/lib/python${PYTHON_MAJOR_MINOR_VERSION}/site-packages/airflow/config_templates/default_airflow.cfg
 
 # Create logs directory, so we can own it when we mount volumes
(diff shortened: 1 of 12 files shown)


kaxil (commented):
I know we purposely ran airflow sync-perm separate in entry point because otherwise the permissions are sync'd in each gunicorn worker.

This created delayed Webserver start times and issues in Azure afaik.

We had discussed this late last year 

cc @ashb  any thoughts?

ashb:
> Because of the custom fab security manager, permissions need to be synced in Airflow on webserver startup

Our security manager doesn't change anything about the permission sync or not -- it only adds a "before_request" filter to create users from the houston-signed JWT token. Also the security manager is only actually enabled in the platform -- the AC images have the security manager installed, but it is dormant.

What is the actual error we are trying to solve here? Do we have a stack trace? Can we have reproduction steps?

Yeah, having this one means that, at webserver start-up _each_ of the 4 worker processes tries to sync permissions, causing massive disk/cpu thrashing and a big spike of load on the DB.

Overriding the entrypoing in an image is going to break _most_ docker images, so anything that recommends doing that is probably wrong 🙁

cmarteepants:
Closing as @kaxil @ashb mentioned slow startup times as this setting syncs everytime the gunicorn worker restarts (which I didn't know, I interpreted the config differently). 

Issue will remain open

PR #223, opened 2021-03-12
Remove slack approvals for checks for other branches

ernest-kr:
Currently when a PR is created, the circleci job is triggered but slack approvals block running the complete build process. This is making the reviewer difficult to check the status. This PR will remove the slack approvals while checking the PRs from other branches. 
Now slack approvals will run only on master.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -68,11 +68,19 @@ workflows:
               }
           requires:
             - static-checks
+          filters:
+            branches:
+              only:
+                - master
       - pause_workflow:
           name: Need-Approval-1.10.5
           requires:
             - Slack_Notification-1.10.5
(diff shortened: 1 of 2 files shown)


kaxil:
I was actually thinking the opposite @ernest-kr where we only have Slack approvals on PRs to only runs tests for versions that have changed but Master runs everything, WDYT?

ernest-kr:
Actually the problem is: the one who is creating the PR does not know if it needs a slack approval. Also the reviewer might want to check if the changes committed affected all the versions of the airflow. 

Also we dont want to unnecessarily push the docker images for all versions especially when we know for certain what airflow release version we want

kaxil (approved):

PR #241, opened 2021-04-12
Fix the builds for 1.10.5

ashb:
The version of flask-admin didn't even _load_ -- meaning migrations wouldn't run.

The removal of packages.lock etc is to avoid false-positives of
vulnerabilities that only apply at build-time, not run time

(We may not release another 1.10.5 image, but we should be _able_ to.)

### DIFF
--- a/1.10.5/buster/Dockerfile
+++ b/1.10.5/buster/Dockerfile
@@ -128,7 +128,7 @@ RUN pip install "${AIRFLOW_MODULE}" \
 RUN cd usr/local/lib/python${PYTHON_MAJOR_MINOR_VERSION}/site-packages/airflow/www_rbac \
   && npm install \
   && npm run prod \
-  && rm -rf node_modules
+  && rm -rf node_modules package.json yarn.lock package-lock.json
 
 
 ################################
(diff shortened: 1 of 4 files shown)


kaxil (approved):

1.10.5/buster/Dockerfile
```diff
@@ -128,7 +128,7 @@ RUN pip install "${AIRFLOW_MODULE}" \
 RUN cd usr/local/lib/python${PYTHON_MAJOR_MINOR_VERSION}/site-packages/airflow/www_rbac \
   && npm install \
   && npm run prod \
-  && rm -rf node_modules
+  && rm -rf node_modules package.json yarn.lock packages.lock
```
kaxil, line 131: We should remove `package-lock.json` too
```suggestion
  && rm -rf node_modules package.json yarn.lock packages.lock package-lock.json
```

kaxil: https://app.circleci.com/pipelines/github/astronomer/ap-airflow/1434/workflows/2e5ad218-b0f3-4c2f-86c0-399c99f20838/jobs/35694

kaxil (approved):

1.10.5/buster/Dockerfile
ashb: `packages.lock` is not a real file -- `package-lock.json` is what I _meant_ to put.

PR #245, opened 2021-04-19
Add 2.0.0-5.dev

kaxil:
This commit adds 2.0.0-5.dev image from
https://github.com/astronomer/airflow/commits/v2-0-0


ernest-kr (commented):
@kaxil why do we need a 2.0.0-5 ?

ernest-kr (approved):

PR #247, opened 2021-04-23
Fix static check

kaxil:
This was introduced in https://github.com/astronomer/ap-airflow/pull/246


ernest-kr (approved):

danielhoherd:
LOOOL, #206 wasn't good enough huh? 😂

PR #249, opened 2021-04-26
Remove AC 1.10.5

kaxil:
Removing AC 1.10.5 since the support for it has ended


ernest-kr (approved):
Dont see 1.10.15 anymore

PR #271, opened 2021-06-05
Create a new Regression Tests Workflow to the Main workflow

ernest-kr:
Create a new Regression Tests Workflow to the Main workflow

We have modified the workflow 
1. Create a new function to check if the version of the airflow is > 2.0
2. Added new workflow regression that is independent of the certified workflow
3. For each airflow 2.0+ version, a new regression workflow is created 
4. Each Regression workflow would contain a first job - called waiter job - this would wait for each airflow version push job in the certified airflow workflow to complete
5. If the waiter job succeeds , we have multiple jobs (Each Set below runs for the three executors)
               1. qa-ui
               2. qa-core-dags
               3. qa-rest-api
               4. qa-dag-params
               5. qa-extended-dags
               6. qa-k8s       
 6. Added code in the push job to write the latest tag of the airflow version into the Database.


ernest-kr:
I might have to get the latest changes from the master again because of the new changes. Meanwhile if you have time can you review this PR?

.circleci/generate_circleci_config.py
```diff
@@ -68,6 +69,12 @@ def main():
         circle_ci_config_file.write(config)
         circle_ci_config_file.write("\n")
 
+def checkair2(airflow_version):
+    airflow2_version_found= re.search(r'^2',airflow_version)
+    if airflow2_version_found is None:
+        return False
+    else:
+        return True
```
kaxil, line 77:
```suggestion
def is_airflow_2(airflow_version):
    return airflow_version.startswith("2.")
```

.circleci/generate_circleci_config.py
```diff
@@ -55,6 +55,7 @@ def main():
     template_env = Environment(loader=FileSystemLoader(searchpath=circle_directory), autoescape=True)
     template_env.filters['dev_releases'] = dev_releases
     template_env.filters['get_airflow_version'] = get_airflow_version
+    template_env.filters['checkair2'] = checkair2
```
kaxil, line 58:
```suggestion
    template_env.filters['is_airflow_2'] = is_airflow_2
```

and use `is_airflow_2` in the config file as it better in terms of naming

PR #277, opened 2021-06-24
Improvements in Providers Repackage Script

kaxil:
- Make `version` optional
- Allow repackaging wheels from local directory
- Don't repackage Kubernetes provider from a URL as it won't have Istio patch.


kaxil:
Synk errors are because of rate limits

ashb (approved):

PR #278, opened 2021-06-24
Update Constraints for AC 2.1.1-1 dev image

kaxil:
New versions of providers have been released which should be used with AC 2.1.1-1 dev image.

This PR/commit updates the constraints to match with https://github.com/apache/airflow/blob/constraints-2-1/constraints-3.7.txt and updates constraints to use the ones for AC


ashb (approved):

ashb:
Do we need something similar for 2.1.0?

ashb:
I guess we don't _need_ it.

kaxil:
> I guess we don't _need_ it.

Yeah since 2.1.0 patches are already out, we shouldn't change deps. Customers can override if they need, but otherwise there will be unknown surprises

PR #279, opened 2021-06-24
Add Env Var for Snowflake Partner Program

kaxil:
Support was added in Snowflake provider: https://github.com/apache/airflow/pull/16420 to attribute usage to Partners.

This PR/commit will adds this Env var for Astronomer <-> Snowflake partnership


ryw:
@paolaperaza @kaxil will ac-runtime (gen2) inherit this setting as well? 
or do we need to apply it in that project as well?

ashb (approved):

kaxil:
> @paolaperaza @kaxil will ac-runtime (gen2) inherit this setting as well?
> or do we need to apply it in that project as well?

Yup, we will apply there too (separately)

PR #289, opened 2021-08-10
Update pip to latest version in buster images

### LINKED ISSUE #288: Update pip version installed in images
As discussed [here](https://github.com/apache/airflow/issues/17453), older versions of pip cause sqlalchemy to update beyond what we want when apache-airflow-providers-snowflake is installed. There is a workaround to pair apache-airflow-providers-snowflake with a restraint of sqlalchemy at the same time, but it would be simpler to just have pip figure it out.

collinmcnulty:
**What this PR does / why we need it**: fixes #288. As far as I can tell, the pip version is simply the latest version at the time that we needed to start specifying it explicitly. Now that we have an error caused by pip's handling of one of the most common provider packages (snowflake) I am trying to update it. I'm basing this fix in large part on [Jed's conclusion](https://github.com/apache/airflow/pull/17452) that the root cause of the snowflake problem is the pip resolver.

**Special notes for your reviewer**: This is my first non-documentation PR for astronomer. I didn't see a way to test if the update to pip would break functionality unless it failed to build. If I'm breaking the code or etiquette, please let me know; I want to help and not hinder.

#### Checklist
- [X] If changing an image, add applicable test in .circleci/test-airflow-image.py


collinmcnulty:
It appears that the build-time-pip-constraints for some images do not actually represent a compatible set of packages. I don't know if that's known/intentional or if we were using pip 19 specifically to avoid this.

jedcunningham:
Those constraints must be built somehow, but a quick look around came up empty. @kaxil, how are those generated?

jedcunningham:
Another thought, we might entertain only updating the pip version for 2.0.0+.

kaxil:
Yup, I would also suggest that. We might also have to think through this as the new PIP version won't allow for inconsistencies at all. I am afraid some users might be relying on "inconsistencies" if they are providing extra packages, we will need to document and tell our CRE team too when new images are released with this change.

collinmcnulty:
@kaxil Happy to look into it. What I'm missing though is which top level packages we truly need and which ones are just dependencies of those top level needs. Since every single package is version-pinned, its harder to determine that. I can look into it either way and just find the minimum compatible environment, but if we have a list of what those top level packages are, that would help me out.

collinmcnulty:
After limiting to airflow 2+, only 2.0.2 had an inconsistent environment.

I have updated the constraints using essentially the method @kaxil described above. I did the pip environment creation without a constraint file then output the pip freeze to make a new constraint file. I removed some packages from the constraint that seemed to have been deliberately removed before (e.g. airflow itself). 

This still leaves the judgment call of if the changed packages are considered "breaking" changes. I don't have a good mechanism by which to determine that. If there are some that changed which we want to avoid changing, I can easily redo this process with a small constraint file of top level items we know we want to remain the same.

collinmcnulty:
And I'm on the CRE team so I can let everybody know about this change if it's accepted.

collinmcnulty:
@kaxil 
> I am afraid some users might be relying on "inconsistencies" if they are providing extra packages

If I'm reading [this](https://pip.pypa.io/en/latest/user_guide/#changes-to-the-pip-dependency-resolver-in-20-3-2020) correctly, I don't think this will cause a problem unless users are relying on inconsistencies within packages that they themselves have provided in the requirements.txt. Conflicts between requirements.txt packages and those we specified in the base image will be ignored and the requirements.txt packages will "win". 

The new resolver still only considers conflicts within a single pip command, so it will happily create an inconsistent environment over the course of several pip installs. Since the requirements.txt is added in a separate pip command, it should be ok, unless I'm missing your point.

2.0.2/buster/build-time-pip-constraints.txt
```diff
@@ -1,87 +1,88 @@
...
-apache-airflow-providers-amazon==1!1.3.0
+apache-airflow==1!2.0.2+astro.4
+apache-airflow-providers-amazon==1!1.4.0
 apache-airflow-providers-celery==1!1.0.1
 apache-airflow-providers-cncf-kubernetes==1!1.2.1
 apache-airflow-providers-elasticsearch==1!1.0.4
-apache-airflow-providers-ftp==1!1.0.1
-apache-airflow-providers-google==1!2.2.0
+apache-airflow-providers-ftp==1!1.1.0
+apache-airflow-providers-google==1!3.0.0
```
kaxil, line 12: This is a breaking change, bumps a major version

2.0.2/buster/build-time-pip-constraints.txt
```diff
@@ -1,87 +1,88 @@
 # Constraint the version pip will install during docker build
 
 adal==1.2.7
-alembic==1.5.8
+alembic==1.6.5
 amqp==2.6.1
-apache-airflow-providers-amazon==1!1.3.0
+apache-airflow==1!2.0.2+astro.4
```
kaxil, line 6: We don't need this as we already have:

https://github.com/astronomer/ap-airflow/blob/7260fe59df278d56d8d95b72f8ae1af84e4e8d92/2.0.2/buster/Dockerfile#L161-L162

2.0.2/buster/build-time-pip-constraints.txt
collinmcnulty: Updated with pinned google provider of 2.2.0

2.0.2/buster/build-time-pip-constraints.txt
collinmcnulty: This has now been removed

.circleci/bin/test-airflow-image.py
```diff
@@ -85,6 +85,9 @@ def test_version(webserver, docker_client):
         post_fix_version_astro = ac_version.rsplit('-')[-1]
     assert post_fix_version_astro == ac_version_postfix_output, \
         f"Incorrect post-fix version in {ac_version_output}"
+    if airflow_2:
+        pip_ver = webserver.check_output("pip --version")
```
kaxil, line 89:
```suggestion
        pip_ver = webserver.pip_package.get_packages()["pip"]
```

.circleci/bin/test-airflow-image.py
```diff
@@ -85,6 +85,9 @@ def test_version(webserver, docker_client):
...
+        assert "21.2.3" in pip_ver
```
kaxil, line 90:
```suggestion
        assert semantic_version(pip_ver) >= semantic_version('21.2.3')
```

kaxil (approved):
Minor suggestion

2.0.0/buster/Dockerfile
```diff
@@ -45,7 +45,7 @@ ARG PYTHON_BASE_IMAGE
 ENV PYTHON_BASE_IMAGE=${PYTHON_BASE_IMAGE}
 ARG PYTHON_MAJOR_MINOR_VERSION
 ENV PYTHON_MAJOR_MINOR_VERSION=${PYTHON_MAJOR_MINOR_VERSION}
-ARG PIP_VERSION="19.3.1"
+ARG PIP_VERSION="21.2.3"
```
kaxil, line 48: `21.2.4` is now available too so probably let's update it to that

kaxil:
```suggestion
ARG PIP_VERSION="21.2.4"
```

2.0.2/buster/Dockerfile
```diff
@@ -45,7 +45,7 @@ ARG PYTHON_BASE_IMAGE
 ENV PYTHON_BASE_IMAGE=${PYTHON_BASE_IMAGE}
 ARG PYTHON_MAJOR_MINOR_VERSION
 ENV PYTHON_MAJOR_MINOR_VERSION=${PYTHON_MAJOR_MINOR_VERSION}
-ARG PIP_VERSION="19.3.1"
+ARG PIP_VERSION="21.2.3"
```
kaxil, line 48:
```suggestion
ARG PIP_VERSION="21.2.4"
```

2.1.1/buster/Dockerfile
```diff
@@ -45,7 +45,7 @@ ARG PYTHON_BASE_IMAGE
 ENV PYTHON_BASE_IMAGE=${PYTHON_BASE_IMAGE}
 ARG PYTHON_MAJOR_MINOR_VERSION
 ENV PYTHON_MAJOR_MINOR_VERSION=${PYTHON_MAJOR_MINOR_VERSION}
-ARG PIP_VERSION="19.3.1"
+ARG PIP_VERSION="21.2.3"
```
kaxil, line 48:
```suggestion
ARG PIP_VERSION="21.2.4"
```

2.1.0/buster/Dockerfile
```diff
@@ -45,7 +45,7 @@ ARG PYTHON_BASE_IMAGE
 ENV PYTHON_BASE_IMAGE=${PYTHON_BASE_IMAGE}
 ARG PYTHON_MAJOR_MINOR_VERSION
 ENV PYTHON_MAJOR_MINOR_VERSION=${PYTHON_MAJOR_MINOR_VERSION}
-ARG PIP_VERSION="19.3.1"
+ARG PIP_VERSION="21.2.3"
```
kaxil, line 48:
```suggestion
ARG PIP_VERSION="21.2.4"
```

kaxil:
Can you rebase on master please -- it should fix the failing test

.circleci/bin/test-airflow-image.py
```diff
@@ -85,6 +85,9 @@ def test_version(webserver, docker_client):
         post_fix_version_astro = ac_version.rsplit('-')[-1]
     assert post_fix_version_astro == ac_version_postfix_output, \
         f"Incorrect post-fix version in {ac_version_output}"
+    if airflow_2:
+        pip_ver = webserver.pip_package.get_packages()["pip"]
+        assert semantic_version(pip_ver) >= semantic_version('21.2.3')
```
kaxil, line 90:
```suggestion
        assert semantic_version(pip_ver['version']) >= semantic_version('21.2.3')
```

kaxil:
Thanks @collinmcnulty

PR #299, opened 2021-08-27
Bump alpine version to `3.14` from `3.10`

kaxil:


pgvishnuram:
@kaxil - can we make this change , 3.14 has most vulnerabilities fixed or 3.15

kaxil:
> @kaxil - can we make this change , 3.14 has most vulnerabilities fixed or 3.15

Since 1.10.7 and 1.10.10 has reached EOL and 1.10.12 will do the same on 31 Jan, I think we won't do this as buster/bullseye images are already available and alpine have been stopped for a while now

PR #316, opened 2021-10-12
Ignore False Positive CVE: `CVE-2020-1747`

kaxil:
Same as #313 but for all the AC versions. Only `pyyaml< 5.3.1` versions are affected, all our images contain `>=5.3.1`


kaxil:
I am starting to hate Trivy, the CVE clearly states :

>A vulnerability was discovered in the PyYAML library in versions before 5.3.1, where it is susceptible to arbitrary code execution when it processes untrusted YAML files through the full_load method or with the FullLoader loader. 

It still show `5.3.1` version as vulnerable 🤦

PR #317, opened 2021-10-12
Rename tag for 2.2.0 docker image to bullseye

kaxil:
This was an oversight, we should have released image with tag `2.2.0-1-bullseye-onbuild` not `2.2.0-1-buster-onbuild`


ashb:
I think if we are changing it we should probably just change it to `debian`

kaxil:
> I think if we are changing it we should probably just change it to `debian`

Correct me if I am wrong but historically we made our tags similar to the Python docker images (`python:3.7-alpine3.10` and `python:3.7-buster` as example). If we follow suite (`python:3.9-bullseye`), `-bullseye` might be better.

Do you have strong opinion on `debian` @ashb or just preference?

PR #340, opened 2021-11-03
Run pre-commit hooks accessing file serially

kaxil:
We had a race condition where the pre-commit hooks were accessing the files parallel-ly mainly for `verify-changelog-entries` and `update-dockerfiles` which caused empty files as `.read()` would entry file if the file is not closed.

This PR fixes it.

Example failure: https://app.circleci.com/pipelines/github/astronomer/ap-airflow/1969/workflows/7d9e8477-6a1c-4960-bf01-c03cf9394af5/jobs/47416


blag (approved):
Yes, thank you for finding and fixing this!

PR #342, opened 2021-11-05
Fix `CVE-2021-21240` and `CVE-2021-33503` in 2.1.0

kaxil:
Fixes `CVE-2021-21240` and `CVE-2021-33503`


jedcunningham (approved):
Should this be on the changelog too?

PR #326, opened 2021-10-26
CircleCI dynamic configuration

blag:
**What this PR does / why we need it**:

* [x] Rework the CircleCI config to autogenerate all of the workflow steps to build all of the images.
* [x] ~Simplify autogeneration to simply read the directory structure to dynamically generate CircleCI config.~
* [x] Split up the pre-commit checks into separate scripts.

**Special notes for your reviewer**:

* I updated the GitHub pull request template, so please review that and make sure I didn't remove any important verification steps.

**Possible Future TODO**:

* [ ] We might also want to look into using CircleCI's native build matrixes to build things, but that might necessitate larger changes to this repo.
* [ ] Create a template directory for new AC releases instead of copying from the latest/closest version.
* [ ] Simplify autogeneration to simply read the directory structure to dynamically generate CircleCI config. Note that we link to individual release changelogs from the `README.md` in this repo, as well as other places, so we'll need to take care not to break those links.

PR #346, opened 2021-11-10
Fix the nightly Job

kaxil:
The Nightly builds were broken when we refactored CircleCi configs: https://github.com/astronomer/ap-airflow/pull/326

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,7 +11,16 @@ workflows:
   setup:
     jobs:
       - generate-config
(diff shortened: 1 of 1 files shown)


blag:
Should we remove the (broken) nightly configuration from `.circleci/config.yml.j2`?

blag (approved):
Not sure how this will interact with the dynamically generated nightly build triggers, but we can find out and fix if it does something weird.

PR #354, opened 2021-11-16
Insert latest tags into a cloud DB and notify on slack

ernest-kr:
Insert latest airflow dev tags into a cloud DB so the automation framework can consume

Also everytime the team has to either go into quay or the actual pipeline to check for new tags. This will provide a slack notification whenever there is a new dev build. 

The latest commit moves the slack notification into a new job

### DIFF
--- a/.circleci/config.yml.j2
+++ b/.circleci/config.yml.j2
@@ -389,7 +389,7 @@ jobs:
           prod_docker_repo_quay_io: "<< parameters.prod_docker_repo_quay_io >>"
           dev_docker_repo_quay_io: "<< parameters.dev_docker_repo_quay_io >>"
       - slack/notify:
-          branch_pattern: master
+          branch_pattern: master,push_tags_db
           channel: qa-team
           custom: |
             {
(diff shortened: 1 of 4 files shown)


.circleci/config.yml.j2
```diff
@@ -534,6 +613,46 @@ commands:
...
+
+                  insert_sql="INSERT INTO tag_version values('${release}','${tag}','${PREVIOUS_TAG}')"
+                  # Connect to the database, run the query, then disconnect DB
+                  PGPASSWORD="$QA_RELEASEINFO_PASSWORD" psql -t -A \
+                  -h "$QA_RELEASEINFO_HOST" \
+                  -p "$PORT" \
+                  -d "$QA_RELEASEINFO_DATABASE" \
+                  -U "$QA_RELEASEINFO_USERNAME" \
+                  -c "$insert_sql"
+                fi
```
kaxil, line 653: This neds to be in a different job

ernest-kr: Kaxil, if I add this to another job, the build number will be change from what we tag for the docker image. It will become difficult to pass on the dynamic value again.
If it is possible can you please suggest?
cc: @kaxil

kaxil: @blag (Drew Hubl) Can help you with that

PR #355, opened 2021-11-16
Add script for building images with different base python versions

collinmcnulty:
Some customers have requested 3.7 versions of the 2.2 image. Kaxil shared instructions for making such an image yourself, and this script codifies and generalizes that process.

I don't know if it's a good idea to make this script part of this repo. If it isn't, I can create a separate repo and give customers more instructions. The idea is very much that this is a use-at-your-own-risk script and I don't want to undercut that by placing it too prominently.

#### Checklist
None of the sections are relevant to this change.

### DIFF
--- a/scripts/build-alt-py-version.sh
+++ b/scripts/build-alt-py-version.sh
@@ -1,0 +1,37 @@
+#!/bin/sh
+
+# This script is intended to be run from the root of ap-airflow
+
+if [ "$#" -ne 2 ]; then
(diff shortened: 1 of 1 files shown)


kaxil (commented):
Looks like we have public-facing docs about it too so probably not worth a script?: https://www.astronomer.io/docs/cloud/stable/develop/customize-image#build-with-a-different-python-version

collinmcnulty:
Ah, I did not see those docs.

PR #361, opened 2021-11-26
Fix bugs with edge builds

kaxil:
Fixes:
- The git push after rebase was missing `--force`
- The build.json step was not persisting the data for the other job to use


.circleci/config.yml.j2
```diff
@@ -2,208 +2,25 @@ version: 2.1
 
 workflows:
   rebase-astro-main:
-    when:
-      and:
-        - equal: [ scheduled_pipeline, << pipeline.trigger_source >> ]
-        - equal: [ "every-midnight-utc", << pipeline.schedule.name >> ]
     jobs:
       - rebase-astro-main
-  certified-airflow:
```
ashb, line 11: Where di this go? It's a very big diff for a small sounding change.

kaxil: My bad.. missed a commit, please check now

ashb (approved):

PR #362, opened 2021-11-29
Emit additional info for debugging nightlies

blag:

Emits additional information that will be useful when debugging nightly builds. Nightly workflows are somewhat difficult to test since they aren't run on PRs and aren't run super frequently, so having good/useful debug info should help surface issues when troubleshooting failed builds.


kaxil (approved):

blag:
Thanks!

PR #380, opened 2021-12-21
Bump test requirements

kaxil:
Bumps some olds requirements


blag (approved):
LGTM. All but one scan job failed, which is weird but probably not related to this.

PR #385, opened 2021-12-22
Fix bash to properly fix failing master

kaxil:
https://stackoverflow.com/a/19670094


jedcunningham (approved):

blag (commented):
Also on line 820.

kaxil:
> Also on line 820.

That one should be OK as `<<parameters.dev_release >>` is out of `[[``

```
elif [[ "$tag" == "<< parameters.tag >>-$CIRCLE_BUILD_NUM" ]] || <<parameters.dev_release >> ; then
```

PR #387, opened 2021-12-22
Better Jinja templates and pre-commit checks

blag:

This PR breaks up the single Jinja template file and simply `include`s the new templates. Each of the new templates is meant to be a specific syntax. That change alone should make it easier to see syntactical errors when editing text in text editors that do syntax coloring (which is pretty much all of them).

This PR also adds pre-commit syntax checks on the new template files, as well as a YAML syntax check on the final generated `config.yml` file as a whole, and syntax checks on the specific sections that are different syntax within the final generated `config.yml` file.

I'm not 100% sure that the pre-commit checks will be super useful. They all require either `jq`, `yq`, or both, and all of the syntax-specific checks are at least somewhat brittle in their own ways. So if these checks end up getting annoying, I wouldn't be opposed to ripping them out.


.pre-commit-config.yaml
```diff
@@ -33,3 +33,66 @@ repos:
...
+        name: Check for workable jq
+        language: system
+        entry: |
+          bash -c "which jq"
+      - id: check-for-yq
+        name: Check for workable yq
+        language: system
+        entry: |
+          bash -c "which yq"
+
```
kaxil, line 47:
```suggestion
```

duplicate?

kaxil (commented):
Waiting for the change to use `shellcheck` instead of `bash -n`

kaxil (approved):

PR #388, opened 2022-01-03
Copy logging sidecar entrypoint to 2.2.3 image

thmzlt:
**What this PR does / why we need it**: Copy logging sidecar entrypoint to 2.2.3 image

**Special notes for your reviewer**: These are the changes suggested by @rob-1126 in his implementation of the sidecar logging feature. I copied them verbatim.

Related to astronomer/issues#3563.

### DIFF
--- a/2.2.4/bullseye/include/entrypoint
+++ b/2.2.4/bullseye/include/entrypoint
@@ -1,52 +1,108 @@
 #!/usr/bin/env bash
-set -e
+set -ex
 
-if [[ $UID == "${ASTRONOMER_UID:-1000}" ]]; then
+# Optional sidecar logging keepalive support
+if [[ -n "${SIDECAR_LOGGING_KEEPALIVE_PORT}" ]] && [[ "$COMPONENT" =~ ^(scheduler|webserver|worker)$ ]] && [[ $1 != "python" ]] ; then
+  SIDECAR_LOGGING_KEEPALIVE_SOCKET=/dev/tcp/127.0.0.1/${SIDECAR_LOGGING_KEEPALIVE_PORT}
+  echo "Connecting to sidecar logging keepalive-daemon on ${SIDECAR_LOGGING_KEEPALIVE_SOCKET}"
+  # loop while we wait for the socket to come up, quashing stderr in the process
+  # once we do manage to make a connection, redirect it to fd #7
+  while true ; do
+    set +e
+    exec 4>&2 2>/dev/null 7<>"$SIDECAR_LOGGING_KEEPALIVE_SOCKET" 2>&4
+    set -e
+    # shellcheck disable=SC2181
+    if [ $? -eq 0 ] ; then
+        echo "Connected to sidecar logging keepalive-daemon on ${SIDECAR_LOGGING_KEEPALIVE_SOCKET}"
+    break
+    else
+      sleep .05
+    fi
+  done
+fi
+
+# Optional support for shipping unstructured logs
+if [[ -n "$UNSTRUCTURED_LOGGING_HOST" ]] && [[ "$COMPONENT" =~ ^(scheduler|webserver|worker)$ ]] && [[ $1 != "python" ]] ; then
+  if [[ -z "${COMPONENT}" ]] ; then
+    echo "COMPONENT must be set when using sidecar logging."
+    exit 2
+  fi
+  # shellcheck disable=SC2153
+  UNSTRUCTURED_LOGGING_SOCKET="/dev/tcp/${UNSTRUCTURED_LOGGING_HOST}/${UNSTRUCTURED_LOGGING_PORT}"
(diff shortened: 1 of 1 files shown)


2.2.3/bullseye/include/entrypoint
```diff
@@ -50,3 +100,9 @@ fi
 
 # Run the original command
 exec "$@"
+
+
+# If we are using a sidecar logging keepalive socket hang up
+if [[ ! -z ${SIDECAR_LOGGING_KEEPALIVE_SOCKET+x} ]] && [[ "$COMPONENT" =~ ^(scheduler|webserver|worker)$ ]]; then
+  echo "goodbye" >> $SIDECAR_LOGGING_KEEPALIVE_SOCKET
+fi
```
danielhoherd, line 108: This appears to be not run any time there is any content in `"$@"`. Tested via:

```
$ cat foo.sh
#!/usr/bin/env bash
echo hi
exec "$@"
echo bye
$ ./foo.sh
hi
bye
$ ./foo.sh whoami
hi
danielh
```

2.2.3/bullseye/include/entrypoint
```diff
@@ -1,5 +1,55 @@
...
+    echo "COMPONENT must be set when using sidecar logging."
+    exit 2;
+  fi
+  UNSTRUCTURED_LOGGING_SOCKET=/dev/tcp/${UNSTRUCTURED_LOGGING_HOST}/${UNSTRUCTURED_LOGGING_PORT}
+  echo "Connecting to tcp logging daemon on ${UNSTRUCTURED_LOGGING_SOCKET}"
+  while true; do
+    set +e
+    # create a pointer to the origianl stdout and stderr
+    exec 3>&1 4>&2 2>&4
+    # clobber old stdout and stderr
```
danielhoherd, line 37: Is this supposed to have a line with some redirects below it?

kaxil:
Don't we need it for other images as well, or are we intentionally targeting 2.2.3 and above?

thmzlt:
> Don't we need it for other images as well, or are we intentionally targeting 2.2.3 and above?

@kaxil, that question came up on our conversation yesterday. Rob initially added support for logging sidecar on 2.1.3 and I did 2.2.3 because it was the latest version. 

@rob-1126 can you check Daniel's comments when you have a chance?

2.2.3/bullseye/include/entrypoint
rob-1126: I think the connection happens on 36. If it was successful it lets the redirect stand.

kaxil:
> @kaxil, that question came up on our conversation yesterday. Rob initially added support for logging sidecar on 2.1.3 and I did 2.2.3 because it was the latest version.

I don't see it in 2.1.3 though

danielhoherd:
@kaxil Thomaz was waiting until everything was approved before copying it to the other supported versions.

2.2.3/bullseye/include/entrypoint
```diff
@@ -1,5 +1,58 @@
 #!/usr/bin/env bash
 set -e
+set -x
+
+
+# Optional sidecar logging keepalive support
+if [[ -n ${SIDECAR_LOGGING_KEEPALIVE_PORT+x} ]] && [[ "$COMPONENT" =~ ^(scheduler|webserver|worker)$ ]] && [[ $1 != "python" ]]; then
```
kaxil, line 7: Why +x here?

danielhoherd: That's an old suggested syntax that we should not be using. https://github.com/koalaman/shellcheck/wiki/SC2268

danielhoherd: wait, no. that's a variable substitution that says "If SIDECAR_LOGGING_KEEPALIVE_PORT is defined, print x." i'm not sure why we would use that here. in any case i will fix it.

2.2.3/bullseye/include/entrypoint
danielhoherd: What i mean is there's a comment that says "Clobber old stdout and stderr" but there is nothing below that comment.

rob-1126:
Agreed. That comment doesn't seem to describe what's going on there at all.
I suggest snipping lines 33-45 and reimplementing a retry fail there that's
less terrible. That code is certainly terrible and entirely untested (my
doing). Lines 33-45 look to not serve their intended purpose at all
(definitely my fault). It looks like at some time in the past lines 46-51
probably resided between lines 38 and 39.

basically to do something in bash like:
while true:
  try:
     setup_redirect()
     break
  catch:
    sleep(.05)


#

2.2.3/bullseye/include/entrypoint
danielhoherd: I removed the `exec` so this should work now.

jedcunningham:
(Closing this because it is stale)

danielhoherd:
FTR this approach as abandoned, so this PR wasn't needed anymore.

PR #395, opened 2022-01-11
Remove support for AC 1.10.10

kaxil:
1.10.10 had reached EOL - https://docs.astronomer.io/enterprise/ac-support-policy/#astronomer-certified-lifecycle-schedule


jedcunningham (approved):

blag:
I've pushed a `v1.10.10` branch based on the current `master` (eg: before this change). We might be able to use branch protections to prevent anybody from pushing to it (and possibly from even opening PRs against it).

That way, just in case there's anybody still modifying 1.10.10 themselves, they still have access to that code, however unsupported. And for the rest of us using `master`, we don't have to see or care about `1.10.10` once this is merged.

PR #409, opened 2022-02-17
Change `auth_backend` to `auth_backends`

ephraimbuddy:
Following the change in Airflow https://github.com/apache/airflow/pull/21472,
we update this as well

### DIFF
--- a/.circleci/bin/test-airflow-image.py
+++ b/.circleci/bin/test-airflow-image.py
@@ -287,9 +287,9 @@ def test_airflow_configs(scheduler, docker_client):
 
         assert scheduler.check_output(
             f"cat {config_file_path} | "
-            "grep '^auth_backend' | awk '{print $3}'"
+            "grep '^auth_backends' | awk '{print $3}'"
         ) == "astronomer.flask_appbuilder.current_user_backend", \
-            "[api] auth_backend needs to be set to 'astronomer.flask_appbuilder.current_user_backend' for Platform"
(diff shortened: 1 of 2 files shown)


.circleci/bin/test-airflow-image.py
```diff
@@ -287,9 +287,9 @@ def test_airflow_configs(scheduler, docker_client):
 
         assert scheduler.check_output(
             f"cat {config_file_path} | "
-            "grep '^auth_backend' | awk '{print $3}'"
+            "grep '^auth_backends' | awk '{print $3}'"
```
kaxil, line 290:
```suggestion
            "grep '^auth_backend' | awk '{print $3}'"
```

We will have to keep this `auth_backend` as <2.3.0 versions will still use it and since is just grep it should work fine for 2.3.0+ too

.circleci/bin/test-airflow-image.py
```diff
@@ -287,9 +287,9 @@ def test_airflow_configs(scheduler, docker_client):
...
         ) == "astronomer.flask_appbuilder.current_user_backend", \
-            "[api] auth_backend needs to be set to 'astronomer.flask_appbuilder.current_user_backend' for Platform"
+            "[api] auth_backends needs to be set to 'astronomer.flask_appbuilder.current_user_backend' for Platform"
```
kaxil, line 292:
```suggestion
            "[api] auth_backend(s) needs to be set to 'astronomer.flask_appbuilder.current_user_backend' for Platform"
```

So it is consistent for both versions less than equal to 2.3.0

kaxil (approved):

PR #415, opened 2022-02-23
Enable building main builds on PR

kaxil:
closes https://github.com/astronomer/issues-airflow/issues/167

This PR will allow running "main" builds on PR too so it is easy to dev and test. It allow allows us to re-trigger a pipeline that ran on the `main` branch

### DIFF
--- a/.circleci/config.yml.j2
+++ b/.circleci/config.yml.j2
@@ -33,9 +33,8 @@ workflows:
       {%- for ac_version, distributions in image_map.items() %}
       {%- set airflow_version = ac_version | get_airflow_version -%}
       {%- set airflow_version_wout_dev = airflow_version | replace('.dev', '') -%}
-      {%- set dev_build = "true" if "dev" in ac_version else "false" %}
-      {%- set edge_build = ac_version | is_edge_build -%}
-      {%- if not edge_build %}
+      {%- set dev_build = ac_version | is_dev_release %}
+      {%- set edge_build = ac_version | is_edge_build %}
 
       - pause_workflow:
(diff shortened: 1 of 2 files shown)


blag:
@jedcunningham and I were discussing this a bit and we figured that any dev build should probably skip needing approval. @kaxil What do you think?

kaxil:
> @jedcunningham and I were discussing this a bit and we figured that any dev build should probably skip needing approval. @kaxil What do you think?

agree, will update

kaxil:
> @jedcunningham and I were discussing this a bit and we figured that any dev build should probably skip needing approval. @kaxil What do you think?

Actually I think the change in the PR already takes care of it in the following line, wdyt?

https://github.com/astronomer/ap-airflow/blob/49368f40e1cdd44410df57fa4863cf43b5efeac1/.circleci/config.yml.j2#L143

blag:
That handles the Slack notifications, but the approval task `pause_workflow` on line 39 will still tell CircleCI to require approval.

kaxil:
> ```yaml
> name: Need-Approval-{{ airflow_version }}
> ```

Updated

kaxil:
> That handles the Slack notifications, but the approval task `pause_workflow` on line 39 will still tell CircleCI to require approval:
> 
> ```yaml
>       - pause_workflow:
>           name: Need-Approval-{{ airflow_version }}
>           requires:
>             - Slack-Notification
>           type: approval
>           filters:
>             branches:
>               only:
>                 - master
> ```

Updated

blag:
Alternative: #418.

PR #428, opened 2022-02-28
Drop non-LTS/stable releases

kaxil:
Today is the last day of supporting non-lts releases as described in https://docs.astronomer.io/software/ac-support-policy/#astronomer-certified-maintenance-policy


jedcunningham (approved):
🎉

PR #434, opened 2022-03-05
Only set versions once, while keeping labels

jedcunningham:

We set versions with ARGS multiple times in our Dockerfiles, but we really only need to do it once.

This is just nicer as only one version string needs to get changed during updates.

#### Checklist

- [ ] If a new distribution or new version of Airflow is added, please make sure:
  + [ ] to follow all of the directions in the `README.md`
  + [ ] that there are Dockerfiles in all base image directories
  + [ ] that all of the new or modified Dockerfiles build successfully

### DIFF
--- a/1.10.15/buster/Dockerfile
+++ b/1.10.15/buster/Dockerfile
@@ -111,10 +111,10 @@ RUN apt-get update \
     && rm -rf /var/lib/apt/lists/*
 
 ARG VERSION="1.10.15-6"
-ARG SUBMODULES="async,azure_blob_storage,azure_cosmos,azure_container_instances,celery,crypto,elasticsearch,gcp,kubernetes,mysql,postgres,s3,emr,redis,slack,ssh,statsd,virtualenv"
-ARG AIRFLOW_MODULE="astronomer_certified[${SUBMODULES}]==$VERSION"
 ARG AIRFLOW_VERSION="1.10.15"
 ARG ASTRONOMER_FAB_SECURITY_MANAGER_VERSION="1.8.4"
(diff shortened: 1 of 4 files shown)


1.10.15/buster/Dockerfile
```diff
@@ -132,6 +132,9 @@ RUN pip install "${AIRFLOW_MODULE}" --constraint /tmp/build-time-pip-constraints
     && pip install "https://github.com/astronomer/astronomer-airflow-scripts/releases/download/v0.0.5/astronomer_airflow_scripts-0.0.5-py3-none-any.whl" \
     && pip install "astronomer-fab-security-manager==${ASTRONOMER_FAB_SECURITY_MANAGER_VERSION}"
 
+LABEL io.astronomer.docker.airflow.version="${AIRFLOW_VERSION}"
```
kaxil, line 135: I recall we had found a bug, there might be an issue somewhere where the labels weren’t preserved in a multi stage build

jedcunningham: Ah, must be fixed now then.

kaxil: Might be https://stackoverflow.com/questions/65331437/difference-in-how-docker-labels-are-applied-in-multi-stage-builds-across-platfor but not sure

blag:
Because we use multi-stage builds, I don't know that removing the duplicate `ARG` directives is going to solve this issue.

From the [Docker documentation on `ARG`](https://docs.docker.com/engine/reference/builder/#scope):

> An ARG instruction goes out of scope at the end of the build stage where it was defined. To use an arg in multiple stages, each stage must include the ARG instruction.

I discovered this via this SO answer:
https://stackoverflow.com/a/53682110

PR #438, opened 2022-03-16
Tweaks, fixups, and general cleanup

blag:

* Fixes instructions for adding a new distro
* Documents all of the different types of builds and how they are triggered
* Always tags `main.dev` images
* Adds labels to images containing the CircleCI workflow ID that build the image and a link to the workflow
* Adjusts the suffix of the nightly build tags to `-nightly+YYYYMMDD`
* Uses the list of approvers that's already defined
* Removes (incorrect) tag from the Slack notification

### DIFF
--- a/.circleci/config.yml.j2
+++ b/.circleci/config.yml.j2
@@ -197,10 +197,10 @@ workflows:
           airflow_version: "{{ airflow_version }}"
           {%- if distribution in ["alpine3.10", "buster"] %}
           tag: "{{ airflow_version }}-{{ distribution }}-onbuild"
-          extra_tags: "{{ airflow_version }}-{{ distribution }}-onbuild-${CIRCLE_BUILD_NUM},{{ ac_version }}-{{ distribution }}-onbuild"
+          extra_tags: "{{ ac_version }}-{{ distribution }}-onbuild"
           {%- else %}
           tag: "{{ airflow_version }}-onbuild"
-          extra_tags: "{{ airflow_version }}-onbuild-${CIRCLE_BUILD_NUM},{{ ac_version }}-onbuild"
+          extra_tags: "{{ ac_version }}-onbuild"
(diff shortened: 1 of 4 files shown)


.circleci/config.yml.j2
```diff
@@ -740,6 +754,9 @@ commands:
       edge_build:
         description: "Indicate if this is an edge build"
         type: boolean
+      nightly_build:
+        description: "Indicate if this is a nightly build"
+        type: boolean
```
kaxil, line 759: We could probably add the default of `true` to be on the safer side, so we don't have to edit/update the non-nightlies

kaxil, line 759:
```suggestion
        default: true
```

README.md
```diff
@@ -4,6 +4,26 @@
...
+| `edge` build  | `main.dev`               | Built from the current `main` branch of [astronomer/airflow](https://github.com/astronomer/airflow) |
+| dev build     | `2.2.4-4.dev`            | Development build, released during ap-airflow changes, including pre-releases and version releases  |
+| nightly build | `2.2.4-nightly+20220314` | Nightly builds, regularly triggered by a CircleCI pipeline sometime during the midnight hour UTC    |
+| release build | `2.2.4-4`                | Release builds, triggered by a release PR                                                           |
+
+Note: Edge builds are always development builds
+
+#### Build matrix
+
+| Build         | Nightly                   | Pre-release PR     | Release PR         |
```
kaxil, line 20: This looks awesome, thanks for adding it

kaxil (approved):

.circleci/config.yml.j2
blag: If I understand this correctly, I think setting the default to `true` would force us to update the non-nightlies. Setting it to `false` would let us omit that in the non-nightlies.

PR #447, opened 2022-03-28
Update dependencies to fix CVEs

jedcunningham:
This fixes CVEs for the following:

2.2.4: glibc, mariadb and locales
2.1.4 & 1.10.15: mariadb and gmp

### DIFF
--- a/.circleci/common.py
+++ b/.circleci/common.py
@@ -34,9 +34,9 @@ def get_airflow_version(ac_version):
 
 
 IMAGE_MAP = collections.OrderedDict([
-    ("1.10.15-7.dev", ["buster"]),
-    ("2.1.4-7.dev", ["buster"]),
-    ("2.2.4-4.dev", ["bullseye"]),
+    ("1.10.15-7-dev", ["buster"]),
+    ("2.1.4-7-dev", ["buster"]),
+    ("2.2.4-4-dev", ["bullseye"]),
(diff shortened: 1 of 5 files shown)


.circleci/common.py
```diff
@@ -34,9 +34,9 @@ def get_airflow_version(ac_version):
 
 
 IMAGE_MAP = collections.OrderedDict([
-    ("1.10.15-6", ["buster"]),
-    ("2.1.4-6", ["buster"]),
-    ("2.2.4-3", ["bullseye"]),
+    ("1.10.15-7.dev", ["buster"]),
+    ("2.1.4-7.dev", ["buster"]),
+    ("2.2.4-4.dev", ["bullseye"]),
```
blag, line 39:
```suggestion
    ("1.10.15-7-dev", ["buster"]),
    ("2.1.4-7-dev", ["buster"]),
    ("2.2.4-4-dev", ["bullseye"]),
```

We need to switch to `-dev` instead of `.dev` so the generated version numbers are valid SemVer.

That should fix the tests.

(Apologies, I didn't announce this at all.)

blag (changes requested):
LGTM except the `.dev` vs. `-dev` issue. But my suggested fix should fix the tests and make the version string a valid SemVer.

ashb (approved):

blag (approved):

PR #451, opened 2022-03-30
Tag main builds with Semver compatible Airflow versions

kaxil:
This PR adds `2.3.0-dev` version to main nightly builds. Currently, the nightlies only publishes `main-XXX` tag and `2.3.0.dev` tag which isn't semver compatible. This PR/commit takes care of that as well as publishes an `-onbuild` image for the same.


ashb:
Python module/Airflow versions _aren't_ SemVer though. What is trying to treat them as SemVer?

ashb:
Not related to problem.

PR #477, opened 2022-05-03
Add logs after successful connection to DB in the entrypoint

kaxil:
This PR/commit will aid in debugging and finding the root cause faster when debugging to
 finding why the migration are not running for example.
 Has it connected to the DB or is it just the last log line it has printed.


ashb (approved):

kaxil:
Test failure is unrelated

blag (commented):
:+1: Apologies for the post-merge review.

PR #497, opened 2022-06-15
Add astronomer certified extensions

ephraimbuddy:
This helps us to now use astronomer/airflow-private for 2.3.4+ images.

The change involves having the astronomer theme as a plugin


jedcunningham:
Do we want the version check here, or still in a separate package? I lean toward separate to keep things simple?

.circleci/config.yml.j2
```diff
@@ -77,8 +98,16 @@ workflows:
           image_name: "ap-airflow:{{ airflow_version }}"
           {%- endif %}
           requires:
-            - static-checks
+            - build-extension-package
+            - test-extension-package
+            {%- for airflow_ref in airflow_ref_list %}
```
ashb, line 103: Why do we need to build more than a single version?

2.3.2/bullseye/Dockerfile
```diff
@@ -110,28 +111,25 @@ RUN apt-get update \
...
-
 # Pip install airflow and astro security manager
-RUN pip install "${AIRFLOW_MODULE}" celery flower \
-        --constraint /tmp/build-time-pip-constraints.txt \
-    && pip install "https://github.com/astronomer/astronomer-airflow-scripts/releases/download/v0.0.5/astronomer_airflow_scripts-0.0.5-py3-none-any.whl" \
-    && pip install "astronomer-fab-security-manager==${ASTRONOMER_FAB_SECURITY_MANAGER_VERSION}"
+COPY airflow_wheels/ /tmp/airflow_wheels
+COPY ac_wheels/ /tmp/ac_wheels
+RUN pip install /tmp/airflow_wheels/${AIRFLOW_WHEEL}[$SUBMODULES] celery flower \
+     --constraint https://raw.githubusercontent.com/apache/airflow/constraints-${AIRFLOW_VERSION}/constraints-${PYTHON_MAJOR_MINOR_VERSION}.txt \
```
ashb, line 130: We shouldn't be using the OSS contraints here -- we want to keep contorl over this ourselves.

2.3.2/bullseye/Dockerfile
```diff
@@ -162,7 +158,7 @@ COPY include/pip.conf /etc/pip.conf
 COPY include/pip-constraints.txt /usr/local/share/astronomer-pip-constraints.txt
 
 # Pin apache-airflow version to avoid accidental upgrade
-RUN pip freeze | grep "apache-airflow==" >>  /usr/local/share/astronomer-pip-constraints.txt
+RUN echo "apache-airflow==1!${ASTRONOMER_CERTIFIED_VERSION}" >>  /usr/local/share/astronomer-pip-constraints.txt
```
ashb, line 161: Why this change?

package/astronomer/certified/plugin.py
```diff
@@ -0,0 +1,100 @@
...
+update_check_interval = conf.getint("astronomer", "update_check_interval", fallback=24 * 60 * 60)
+
+
+
+
+class AstronomerCertifiedPlugin(AirflowPlugin):
+
+    name = "astronomer_certified_extensions"
+
+    flask_blueprints = [ACThemeBlueprint()]+[UpdateAvailableBlueprint()] if update_check_interval != 0 else []
```
ashb, line 24: Slightly longer but _much_ easier to read:

```suggestion
    flask_blueprints = [ACThemeBlueprint()]
    if update_check_interval:    
        flask_blueptrints += [UpdateAvailableBlueprint()]
```

package/astronomer/environment.py
```diff
@@ -0,0 +1,3 @@
+import os
+
+ASTRONOMER_CERTIFIED_VERSION = os.environ.get("ASTRONOMER_CERTIFIED_VERSION", "2.3.2+astro.2")
```
ashb, line 3: A file for this is overkil. Just put this one line where it is needed.

Additionally having a fallback version of 2.3.2 seems odd -- saying "unknown" would be better.

2.3.2/bullseye/Dockerfile
ephraimbuddy: Because I installed it from the wheel, the command returns `apache-airflow @ file://..`  so I had to do it this way. Now that we would be building the wheel from airflow-private, I will modify this change

2.3.3/bullseye/build-time-pip-constraints.txt
```diff
@@ -29,9 +29,9 @@ Pygments==2.12.0
 Rx==3.2.0
 SQLAlchemy-JSONField==1.0.0
 SQLAlchemy-Utils==0.38.2
-SQLAlchemy==1.4.9
+SQLAlchemy==1.4.27
```
jedcunningham, line 32: Wait, why do we have these package bumps?

2.3.3/bullseye/include/pip.conf
```diff
@@ -1,3 +1,4 @@
 [global]
+index-url = https://pip.astronomer.io/v2/
 extra-index-url = https://pip.astronomer.io/simple/
```
jedcunningham, line 3:
```suggestion
```

We should get rid of this one while we are at it.

package/astronomer/certified/plugin.py
```diff
@@ -0,0 +1,30 @@
...
+        @functools.wraps(fn)
+        def run_before(*args, **kwargs):
+            pre_fn()
+            fn(*args, **kwargs)
+
+        setattr(mod_or_cls, target, run_before)
+
+    def on_load(cls, *args, **kwargs):
+        # Borrowed concept from version_check plugin.
+        # Seed the log template table before running the scheduler loop
```
jedcunningham, line 25: Can we hook into apply migrations or something instead? Something that is called only during db upgrade?

package/dev-requirements.txt
```diff
@@ -0,0 +1,5 @@
+
+--constraint="https://raw.githubusercontent.com/apache/airflow/constraints-2.3.3/constraints-3.7.txt"
+apache-airflow==2.3.3
+
+-e .[test]
```
jedcunningham, line 5: Do we actually need all this to build the package?

ephraimbuddy: Yeah. This is for testing. I had issues when installing airflow, using the constraint solved it but I will investigate again

2.3.3/bullseye/include/pip.conf
ephraimbuddy: We still have version check and fab security manager at /simple.

jedcunningham: We should republish them on v2 then (or install via url). (extra-index-url is a security concern, part of the driver for this change)

.circleci/config.yml.j2
```diff
@@ -51,7 +56,11 @@ workflows:
           extra_args:
             {#- If you modify this, make sure you also modify it in the 'nightly' workflow -#}
             {%- if not edge_build %}
+            {%- if distribution=='bullseye'%}
+            --build-arg VERSION=$(curl https://pip.astronomer.io/v2/apache-airflow/latest-{{ airflow_version_wout_dev }}.build)
+            {%- else %}
             --build-arg VERSION=$(curl https://pip.astronomer.io/simple/astronomer-certified/latest-{{ airflow_version_wout_dev }}.build)
```
jedcunningham, line 62: Btw, we will want to do this for 2.1.4 also. Probably worth leaving 1.10.15 alone, only have to support it for 2 more months.

ephraimbuddy: Do you mean building 2.1.4 from airflow-private?

jedcunningham:
@ephraimbuddy, can we resurrect this for 2.3.4 and 2.4.1?

ephraimbuddy:
> @ephraimbuddy, can we resurrect this for 2.3.4 and 2.4.1?

I have updated this. We need to update airflow-private with 2.3.4+astro.3. This is currently only in astronomer/airflow repo.

Need reviews

.circleci/common.py
```diff
@@ -35,6 +35,6 @@ def get_airflow_version(ac_version):
 
 IMAGE_MAP = collections.OrderedDict([
     ("2.1.4-9", ["buster"]),
-    ("2.3.4-3", ["bullseye"]),
+    ("2.3.4-2", ["bullseye"]), # TODO: Update this to 2.3.4-3 which is the latest
```
ephraimbuddy, line 38: 2.3.4-3 is not available in airflow private yet

package/astronomer/certified/plugin.py
```diff
@@ -0,0 +1,34 @@
...
+            fn(*args, **kwargs)
+
+        setattr(mod_or_cls, target, run_before)
+
+    @classmethod
+    def on_load(cls, *args, **kwargs):
+        # Borrowed concept from version_check plugin.
+        # Seed the log template table before synchronizing the template
+        version = os.environ.get("ASTRONOMER_CERTIFIED_VERSION", "Unknown")
+        if '2.1.4' not in version:
```
jedcunningham, line 29: Do we need this conditional still?

package/dev-requirements.txt
```diff
@@ -0,0 +1,5 @@
+
+--constraint="https://raw.githubusercontent.com/apache/airflow/constraints-2.3.4/constraints-3.10.txt"
+apache-airflow==2.3.4
```
jedcunningham, line 3: Should this be 2.4.1 now? Does it even matter?

jedcunningham (approved):

ephraimbuddy:
I don't think there's anything we can do about this trivy scan right now for dev. It won't be there on the final image. Here's my simple test:
```python

>>> from packaging.version import Version
>>> x = Version('2.3.4.dev0+astro.1')
>>> y = Version('2.3.4')
>>> x>y
False
>>> y>x
True
>>> k = Version('2.3.4+astro.1')
>>> k>y
True
>>> 
```
When we have `dev`, it will be less than 'rc' and 'main release'

jedcunningham (approved):

PR #500, opened 2022-07-07
Remove providers dependency on astronomer certified

ephraimbuddy:


ashb (commented):
Do we even need to repackage the providers of we aren't changing the dep? This is most (all?) of why did it

ephraimbuddy:
> Do we even need to repackage the providers of we aren't changing the dep? This is most (all?) of why did it

I think we can just be packaging only the providers that we have to cherry-pick code into

jedcunningham (approved):
Yeah, which means there really is only one, `cncf.kubernetes`. We should wait to merge this until we are ready to switch over though.

kaxil:
And keep a close eye on the Elasticsearch provider! I know one of the versions wasn't compatible with the platform so we had pinned the version in constraints - https://github.com/astronomer/ap-airflow/pull/294 -- so we don't need to repackage but just keep an eye out

PR #502, opened 2022-07-11
Automate smoke and regression tests

blag:

This PR attempts to automate smoke and regression tests for QA. It does so by cloning a specific branch of the QA repo, updating a config file with the version of Airflow that was built, commits, and pushes the commit to that repo. The CircleCI workflows implemented on those branches in that repository will handle running the tests and reporting results.

This pushes commit/s to a private repository and may change, which is why the repo name is pulled from an environment variable. And this uses a manually created deploy key to push changes to the private repository.

The branch names, config file name and (presumably) format, and the configured value are all left in cleartext, but that shouldn't be an issue.


.circleci/kickoff_qa_tests.sh
```diff
@@ -0,0 +1,82 @@
...
+if [[ -d "${TEST_REPO_NAME}" ]]; then
+    cd "${TEST_REPO_NAME}"
+    git fetch
+    git reset --hard origin/$(git branch --show-current)
+else
+    git clone ${GIT_REPOSITORY}
+    cd ${TEST_REPO_NAME}
+fi
+
+git branch --show-current
```
blag, line 48: I left this in intentionally so it appears in CircleCI logs. It's one less thing to try to debug without logs.

.circleci/kickoff_qa_tests.sh
```diff
@@ -0,0 +1,100 @@
...
+    git switch ${BRANCH_NAME}
+    git fetch
+    git reset --hard origin/${BRANCH_NAME}
+else
+    git clone --branch ${BRANCH_NAME} ${GIT_REPOSITORY}
+    cd
+fi
+
+git branch
+
```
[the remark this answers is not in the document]
blag: Nope, I want that visible in CircleCI logs, so it's one less thing to debug.

.circleci/bin/test-airflow-image.py
```diff
@@ -220,8 +220,11 @@ def test_airflow_trigger_dags(scheduler):
         dag_state_command = "airflow dag_state example_dag 2020-05-01"
 
     assert "Dag: example_dag, paused: True" in scheduler.check_output(pause_dag_command)
-    assert "Created <DagRun example_dag @ 2020-05-01T00:00:00+00:00: " \
-           "test_run, externally triggered: True>" \
+    # Airflow commit:
+    # https://github.com/apache/airflow/commit/7850dc3c714ef188014a08aae63b4d344242208e
+    # changed the log format, so we relax this test to just test for a short-ish
+    # substring of the log line
```
jedcunningham, line 226: nit: not sure we need the history here.

jedcunningham (approved):

PR #507, opened 2022-07-22
Fix and improve triggering QA tests

blag:

* Kick off QA smoke tests from the `test` branch (so I don't have to commit and push to `master` to test fixes)
* Fix logic that gates kicking off tests for dev builds
* Explicitly check out the code before using it
* Fix the order of arguments to `kickoff_qa_tests.sh`
* Fix checking arguments in `kickoff_qa_tests.sh`
* Improve the logic to update the code from `origin` - fast forward rebases/merges are now supported
* Update the correct file
* Install and use `yq` to update the YAML config files instead of blindly using `sed` - this should be much more robust!
* Configure a git user and email address
* Allow empty QA repo commits - this allows us to force test runs by adding and pushing empty commits
* Add a `DRY_RUN` variable to prevent pushing code, but don't document it
* Better debugging and error messages

This should be the last major code churn to enable QA tests (famous last words...) - the rest of the bugs will need to be worked out in the QA repo.

### DIFF
--- a/.circleci/config.yml.j2
+++ b/.circleci/config.yml.j2
@@ -216,17 +216,19 @@ workflows:
                 - master
 
       {#- Only dev and edge builds kick off smoke and regression tests (for now) #}
-      {%- if (dev_build and airflow_version in dev_allowlist) or edge_build %}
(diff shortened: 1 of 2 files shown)


dstandish (approved):

blag:
Thanks for the review!

PR #509, opened 2022-07-26
Fix JSON syntax and pre-commit config

blag:

This PR fixes a [previous error encountered by CircleCI](https://app.circleci.com/pipelines/github/astronomer/ap-airflow/3378/workflows/87b3ebea-cde3-43cc-821c-97925ec97fd4), specifically [this one](https://app.circleci.com/pipelines/github/astronomer/ap-airflow/3379/workflows/1b6b70a9-3f7c-4281-82ba-2b096d8b09dd). I also figured out how to turn on JSON checks for Jinja templates that render JSON files, so we shouldn't hit that specific issue again in the future.

I tried to ensure that the `push` command is valid Bash, but I couldn't get shellcheck and embedded CircleCI syntax to play nicely together _and_ actually fail on legitimately incorrect Bash syntax. I could easily get it to silently pass when it shouldn't, but I couldn't get it to properly fail while also including the CircleCI `<< parameter.variable_name >>` syntax.

However, in the process of running shellcheck on the repository, I was able to improve a lot of shell-isms in our Bash, so that's still a small win.


.circleci/push.bash.j2
```diff
@@ -5,27 +5,27 @@ export NEW_POINT_RELEASE=true
 
 IFS="," read -ra DOCKER_TAGS \<<< "<< parameters.extra_tags >>"
 
-echo "DOCKER_TAGS: ${DOCKER_TAGS[@]}"
+echo "DOCKER_TAGS: " "${DOCKER_TAGS[@]}"
```
ashb, line 8: Nit:

```suggestion
echo "DOCKER_TAGS: ${DOCKER_TAGS[@]}"
```

.circleci/push.bash.j2
```diff
@@ -5,27 +5,27 @@ export NEW_POINT_RELEASE=true
...
 
-if [[ "<< parameters.nightly_build >>" =~ (True|true) ]]; then
+if [[ "<< parameters.nightly_build >>" ]]; then
```
ashb, line 10: I'm not sure this is correct: _any_ value will pass this:

```bash
$ bash -xc '[[ "false" ]] && echo passed'
+ [[ -n false ]]
+ echo passed
passed
```

.circleci/push.bash.j2
```diff
@@ -5,27 +5,27 @@ export NEW_POINT_RELEASE=true
...
   # Docker tags must match [\w][\w.-]{0,127}
   # https://github.com/distribution/distribution/blob/514cbd71bedb3331f53e2fe5293075726c67e68a/reference/regexp.go#L41
   TAG_PRERELEASE_SUFFIX="-nightly-$(date +%Y%m%d)"
   echo "Adding '${TAG_PRERELEASE_SUFFIX}' to all docker tags"
   for (( i=0; i<{% raw %}${#DOCKER_TAGS[@]}{% endraw %}; i++ )); do
     DOCKER_TAGS[$i]=${DOCKER_TAGS[$i]}${TAG_PRERELEASE_SUFFIX}
   done
-  echo "DOCKER_TAGS->  ${DOCKER_TAGS[@]}"
+  echo "DOCKER_TAGS->  " "${DOCKER_TAGS[@]}"
```
ashb, line 18: Nit

```suggestion
  echo "DOCKER_TAGS->  ${DOCKER_TAGS[@]}"
```

.circleci/push.bash.j2
blag: Shellcheck has [an opinion on concatenating strings and arrays](https://www.shellcheck.net/wiki/SC2145), but I'll change it back. :+1:

.circleci/push.bash.j2
blag: https://www.shellcheck.net/wiki/SC2145

.circleci/push.bash.j2
blag: Good catch! This should work instead:

```bash
if "<< parameters.nightly_build >>"; then
```

```bash
❯ if [[ "false" ]]; then echo yes; else echo no; fi
yes

❯ if "false"; then echo yes; else echo no; fi
no

❯ if "true"; then echo yes; else echo no; fi
yes
```

Fix incoming.

jedcunningham (approved):

.circleci/push.bash.j2
ashb: This only works if it is `true` or `false` as those are commands that bash runs - `True` would fail with "command not found"

blag: Huh, TIL. I didn't think quoted strings would be interpreted as commands. I'll fix this later today.

PR #515, opened 2022-08-10
Stop building from main

blag:

We can stop building from `main` now. 🎉

In an attempt to minimize churn, I didn't change any of the template branching logic, since I don't know that that buys us a whole lot anymore.


jedcunningham (approved):
Still might be worth at least seeing what removing the branching logic would look like? Totally up to you.

ashb (approved):

blag:
I'll keep it in for now, just in case we want to turn it back on. And other folks might also want to build `main`/edge builds, so it can still serve as an example for others.

PR #522, opened 2022-08-29
Mitigate dependency confusion attacks

qequ:
Changes to dockerfile and pip.conf for 1.10.15, 2.14, 2.3.3, 2.3.4 and main to mitigate Dependency confusion attacks.

reference; https://medium.com/@alex.birsan/dependency-confusion-4a5d60fec610

CC: @dgutson  @arieltorti


blag (approved):
LGTM, thanks!

ashb:
This might cause problems at run/ONBUILD time. We'll run some checks

dgutson:
> This might cause problems at run/ONBUILD time. We'll run some checks

ping? :)

dgutson:
Hi @ashb , this is a security issue and therefore we're in a zugzwang situation that leaves us no other alternative that create a fork and file the CVE. We do not want to reach that stage. Do you have an ETA for this?

ashb:
@dgutson @qequ sorry for the silence, I was all hands on deck getting Airflow 2.4.0 out! I'm now looking at this and it's my top priority for now.

I'll get back to you shortly.

1.10.15/buster/include/pip.conf
```diff
@@ -1,3 +1,2 @@
 [global]
-extra-index-url = https://pip.astronomer.io/simple/
```
kaxil, line 2: We can do the following too
```
index-url = https://pip.astronomer.io/simple/
extra-index-url = https://pypi.org/simple
```

ashb: Please see https://github.com/astronomer/ap-airflow/pull/536

arieltorti: This won't solve the security problem, both indexes have the same priority order, see: https://github.com/pypa/pip/issues/5045, https://github.com/pypa/pip/issues/8606

ashb: Can you explain in a bit more detail how this is a problem for these specific repos? Because we're not talking about a "traditional" private repo here

arieltorti: Let's make an example with a specific package: `astronomer-fab-security-manager`

If we install it using `pip install astronomer-fab-security-manager==<version> --extra-index-url https://pip.astronomer.io/simple/` pip will query both indexes and look for the *best match*. The main problem is that index priority is undefined: https://github.com/pypa/pip/issues/8606 meaning no index takes priority over the other and the selection algorithm may change between versions.

A malicious actor could publish a package with the same name and version to https://pypi.org/simple resulting in it being used instead of the one in the astronomer index.

What's more, pip selects packages based on *best match* taking into account [Compability Tags](https://peps.python.org/pep-0425/). Even if one index gets consistently prioritized the malicious actor could just publish the same package with different combinations of tags to make pip select his package over the astronomer one. As an example, running pip on my machines lists 726 compatible tags, with the best match being: `cp310-cp310-manylinux_2_31_x86_64`, in contrast astronomer is publishing packages with a single tag `py3`. 

Summarizing, there's many caveats to how a package gets selected from multiple indexes and the best approach is to restrict private packages to private indexes.

[Hashes](https://pip.pypa.io/en/stable/topics/secure-installs/#hash-checking-mode) can help prevent some of this issues, because if a package doesn't match the hash it won't get downloaded, but this only moves the problem to the left, when someone upgrades or selects the dependency for the first time it *must* make sure it got downloaded from the private registry.

arieltorti: @ashb Did you get a chance to look at this?

dgutson: @ashb Ariel provided actionable information to create an exploit. We are forced to file the CVE as soon as possible.
@arieltorti shall we create an attack PoC? Maybe this way the maintainers will react.
We have enough material to present this in DefCon.

dgutson: We did everything we could as a responsible disclosure.

ashb: Sorry, GitHub notifications is not a reliable way of making sure I see something :(

ashb: I have uploaded https://pypi.org/project/astronomer-fab-security-manager/0.0.0/ and https://pypi.org/project/astronomer-airflow-version-check/0.0.0/ to public PyPi repo, so no one else can register something malicious. This also defends against all currently published versions rather than just future ones.

I'm really sorry for missing your responses, it was not the kind of timely response I wanted to give to security reports!

dgutson: Thanks, just in case we should remove actionable information.
CC @arieltorti

ashb:
Fixed this issue another way (by creating name-squatting packages on pypi)

Also worth noting that we avoided this problem for our newer Astronomer Runtime images which we encourage all users to use for new deployments instead. There we fixed the problem by using a different pip repo which implements the "redirect through" approach, so the _repo_ checks if the package exists locally, and if not will issue an HTTP redirect to the public PyPI repo.

arieltorti:
Awesome @ashb ! That will work as well, we just need to make sure to name-squat all newly added dependencies.

Thanks for all the effort in understanding and fixing the issue

PR #536, opened 2022-09-20
Switch to using `pip.astronomer.io` as primary index

ashb:
And make the default public pip repo the extra/secondary index.

This has the effect that for packages we (re)publish in `pip.a.io` _only_ those versions will be available, but that was the intent all along

Closes #522


kaxil (approved):

jedcunningham (approved):

ashb:
https://github.com/astronomer/ap-airflow/pull/537 should fix the build

ashb:
Not needed, we have reserved the package names on Pypi

PR #540, opened 2022-09-23
Fix nightly scans

blag:

Fix the nightlies. If we can't load images before scans, pull them from the registry. This allows us to scan images that we haven't just built, which is what happens for builds we've already released.

### DIFF
--- a/.circleci/config.yml.j2
+++ b/.circleci/config.yml.j2
@@ -905,7 +905,7 @@ commands:
               --label io.astronomer.repo.commit_sha="${CIRCLE_SHA1}" \
               --label io.astronomer.repo.url="${CIRCLE_REPOSITORY_URL}" \
               --label io.astronomer.ci.build_url="${CIRCLE_BUILD_URL}" \
-              --file '<< parameters.path>>/<< parameters.dockerfile >>' \
+              --file '<< parameters.path >>/<< parameters.dockerfile >>' \
               << parameters.extra_args >> '<< parameters.path >>'
             docker save -o saved-images/<< parameters.image_name >>.tar '<< parameters.image_name >>'
             docker inspect << parameters.image_name >>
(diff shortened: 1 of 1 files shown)


.circleci/config.yml.j2
```diff
@@ -919,7 +921,7 @@ commands:
       - attach_workspace:
           at: {{ workspace_prefix }}
       - run:
-          name: Load archived Airflow Docker image
+          name: Load archived Airflow Docker image or pull it from the registry
```
jedcunningham, line 924:
```suggestion
          name: Load archived Airflow Docker image
```

.circleci/config.yml.j2
```diff
@@ -631,8 +631,10 @@ jobs:
       - attach_workspace:
           at: {{ workspace_prefix }}
       - run:
-          name: Load archived Docker image
-          command: docker load -i {{ workspace_prefix }}/saved-images/<< parameters.image_name >>-onbuild.tar
+          name: Load archived Docker image or pull it from the registry
+          command: |
+            docker load -i {{ workspace_prefix }}/saved-images/<< parameters.image_name >>-onbuild.tar || \
+            docker pull quay.io/astronomer/<< parameters.image_name >>-onbuild
```
jedcunningham, line 637: I'm not sure this will actually work, as it's scanning `"<< parameters.image_name >>-onbuild"`. You'd have to tag it for this to work, and that feels a little messy?

.circleci/config.yml.j2
blag: Fixed, thanks.

.circleci/config.yml.j2
blag: Refactored to something that I think will work better.

.circleci/config.yml.j2
```diff
@@ -464,23 +500,41 @@ workflows:
       # {{ airflow_version }}
       {%- for distribution in distributions %}
       # {{ airflow_version }} - {{ distribution }}
+      - pull-image:
```
jedcunningham, line 503: I just tried scanning an image I don't have pulled and it worked. Do we even need this step?

jedcunningham (approved):

blag:
The Snyk orb had a few troubles there, but it's working now and (aside from scan failures) CI is passing. Merging, and I'll pay attention to this tonight when they run.

PR #549, opened 2022-10-12
add analytics plugin

aliotta:
**What this PR does / why we need it**: Adds an airflow plugin that adds an api that enables houston to pull data on airflow task usage data. 

It also adds a page  linked in the airflow top nav that allows user to hit the endpoint via a form.

## Related Issue(s)
https://github.com/astronomer/issues/issues/5110

#### Checklist


aliotta:
Any thoughts on whether this PR needs tests and what I would test for if so?

aliotta:
I ran 2.1.4 locally and did verify the plugin installed correctly.

2.1.4/buster/Dockerfile
```diff
@@ -131,8 +131,8 @@ COPY build-time-pip-constraints.txt /tmp/build-time-pip-constraints.txt
 RUN pip install "${AIRFLOW_MODULE}" \
         --constraint /tmp/build-time-pip-constraints.txt \
     && pip install "https://github.com/astronomer/astronomer-airflow-scripts/releases/download/v0.0.5/astronomer_airflow_scripts-0.0.5-py3-none-any.whl" \
-    && pip install "astronomer-fab-security-manager==${ASTRONOMER_FAB_SECURITY_MANAGER_VERSION}"
-
+    && pip install "astronomer-fab-security-manager==${ASTRONOMER_FAB_SECURITY_MANAGER_VERSION}" \
+    && pip install https://github.com/astronomer/airflow-analytics-plugin/releases/download/1.0.0/astronomer_analytics_plugin-1.0.0-py3-none-any.whl
```
ephraimbuddy, line 135: Should we publish this plugin in `pip.astronomer.io/simple/`

jedcunningham: For now the simplicity of getting it from github makes sense I think. We can look at automating the releases to the bucket if we end up needing to later.

blag (approved):
If the HTML page of the plugin is going to be visible to end users, it might be a good idea to fix/clean/style it up a bit. But otherwise I think this is okay to merge.

rishkarajgi:
> If the HTML page of the plugin is going to be visible to end users, it might be a good idea to fix/clean/style it up a bit. But otherwise I think this is okay to merge.

@blag Have removed the nav item in the 1.0.2 release and updated the PR with the new release version

blag (approved):

PR #550, opened 2022-10-14
Add changelog for 2.3.4-4-dev

ephraimbuddy:

#### Checklist

- [ ] If a new distribution or new version of Airflow is added, please make sure:
  + [ ] to follow all of the directions in the `README.md`
  + [ ] that there are Dockerfiles in all base image directories
  + [ ] that all of the new or modified Dockerfiles build successfully

### DIFF
--- a/2.3.4/bullseye/trivyignore
+++ b/2.3.4/bullseye/trivyignore
@@ -1,0 +1,4 @@
+# Ignore these CVEs only for the dev image.
+# The release image would pass the CVE scan
+CVE-2022-38054
+CVE-2022-40604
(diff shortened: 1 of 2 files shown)


2.3.4/bullseye/trivyignore
```diff
@@ -0,0 +1,4 @@
+# Ignore these CVEs only for the dev image.
+# The release image would pass the CVE scan
+CVE-2022-38054
+CVE-2022-40604
```
ephraimbuddy, line 4: We will remove these when releasing. Because the airflow version has dev e.g `2.3.4.dev0+astro.4`, which is lesser than `2.3.4` trivy scan fails. We should be good when the version is just `2.3.4+astro.4`

jedcunningham (approved):

